Issue #482, opened 2014-09-29
maildir:/path deliveries


poolpOrg:
It should be possible to declare maildir:/path and mbox deliveries in aliases and ~/.forward just like we can declare files and redirects.

Edit: forget about mbox deliveries.

poolpOrg:
For reference, below is a discussion I had with @snimmagadda on IRC:


11:36 < __gilles> the aliases and forward code are almost identical
11:36 < __gilles> the main problem is the syntax
11:36 < __gilles> two solutions:
11:37 < __gilles> 1- either we need a different syntax than  maildir:/path, which would account for the delivery user, so that:
11:37 < __gilles> gilles.chehade       maildir:~/       has a way to figure out it should be delivered as gilles
11:37 < __gilles> and i have 0 idea how we can do that in a nice way
11:38 < __gilles> 2- or we consider that maildir:/path is only allowed when the key is an expandnode of type username
11:38 < __gilles> so that, you're not allowed to do:
11:38 < __gilles> gilles.chehade       maildir:~/
11:38 < __gilles> but you're allowed to do:
11:38 < __gilles> gilles.chehade      gilles
11:38 < __gilles> gilles      maildir:~/
11:39 < __gilles> where "gilles" HAS to be a local user
11:39 < __gilles> it's the nicest way, but a bit trickier
11:39 < __gilles> and in that case, it would be identical to .forward
11:40 < __gilles> anyways, you should really work on getting it to work in aliases as it will automagically work in .forward

poolpOrg:
Feature committed in master / portable, already part of snapshots.

PR #522, opened 2015-01-09
Implement maildir:/path expansion in aliases and forward file.

snimmagadda:
#482


poolpOrg:
I did not forget this pull request, I'll take care of it tomorrow as some backports from OpenBSD really need to be handled first to make sure we can easily revert them without your merge conflicting.

PR #528, opened 2015-01-28
Backport OpenBSD -current changes.

snimmagadda:
Backport latest commits from OpenBSD cvs tree. Compile tested on OpenBSD -current. The master branch now relies on SSL_CTX_use_certificate_chain(3) that got added post 5.6 release and would no longer compile on OpenBSD 5.6 release.


poolpOrg:
can you please separate these two ?

The 5dfecad one is right and I want to merge it, the other I would like a libressl.c file to be temporarily introduced in git so that we can put these functions in it. This way we can conditionnally compile it in the master branch until we can enforce a LibreSSL dependency.

I agree with Reyk that this is where we want to go, but just removing them and coping with the aftermath is too much of a hassle when we're dealing with OpenBSD stable / current as well as portable altogether and LibreSSL is not widely available yet.

snimmagadda:
Will generate a new pull request for each individual changes.

Issue #343, opened 2013-10-25
discover


poolpOrg:
we should have a smtpctl command to teach the scheduler about new messages in the queue (that could have been moved back by an admin for instance)

poolpOrg:
implemented by @snimmagadda, committed to both master / portable, will be in next snapshot

PR #533, opened 2015-02-01
smtpctl discover subcommand.

snimmagadda:
#343


ericfaurot:
Hi,

Nice diff.  I am wondering whether it is preferable to actually send the discover message to the queue process, which would then trigger a (partial) queue walk and reload the envelope(s), just like at startup. This would also make the feature backend-independent.

poolpOrg:
yes, I agree with eric, having an indirection through the queue and having a partial queue walk would be nice

snimmagadda:
Hello Eric, Gilles,
Thank you for the feedback. Implementing this feature in a backend independent way would indeed be better than this approach. I will come up with a new diff shortly.

Issue #563, opened 2015-05-06
autohell improvement


poolpOrg:
Because our autotools layer was inherited from OpenSSH and extended to build the asr resolver before it became its own project:

1- it contains lots of checks that make no sense (ie: do we support DNSSEC ? SSH ?)
2- it tests a lot of functions / headers that we don't rely upon
3- some checks can't even succeed because the conditions can never be true
4- there is inconsistency in options naming convention (ssl-dir, asr, libevent-dir, zlib)
5- lots of checks are for systems we know we can't/won't run on in a foreseeable future
6- some checks that are identical have slight differences in behaviour
7- there is absolutely no structure in the configure.ac file, all checks are mixed
8- it is plain unreadable and so complex no one wants to actually improve it...

There is no reason we shouldn't apply the same level of quality to our build framework than to the rest of the code, and it has to be done RIGHT NOW before we publish new major release and -extras which will contribute getting this out of hands...

I have created a branch where I have tackled these issues and while I'm not done yet, I have managed to make configure.ac far more readable, bring coherency in how options are processed / named, and more importantly factor the checks so that the m4/ can be shared across several projects, each one keeping only project-specific stuff in configure.ac

I have converted various library checks to macros so that you can basically do:

    CHECK_COMPILER([])    # checks compiler + supported compiler flags / optionss
    CHECK_SYSTEM([])        # check operating system + specific features/limitations
    
    WITH_LIBASR([])        # generate --with-libasr=PATH
    WITH_LIBEVENT([])    # generate --with-libevent=PATH
    
    CHECK_LIBASR([])      # check headers + lib for libasr
    CHECK_LIBEVENT([])  # check headers + lib for libevent
    CHECK_LIBSSL([])

There are many more but you get the idea:

1- WITH_ macros uniformize how we present options
2- CHECK_ macros isolate checks for a specific feature so it isn't spread everywhere
3- CHECK_COMPILER handles all compiler specific checks (options, etc...)
4- CHECK_SYSTEM handls all system specific checks

I will push this further by providing macros to test subset of features at once to shorten the configure.ac file further, the goal being that configure.ac only contains calls to OpenSMTPD-specific tests + calls to generic macros.

When this is done, we can copy m4/ to extras and shorten the configure there too.

This work is done in branch autohell.

poolpOrg:
This will not make the milestone, rather than cleaning up the existing mess, I've started working on a long-term fix for the issue.

PR #564, opened 2015-05-07
Add libasr to CFLAGS and LIBS

evadot:
- Add libasr to LIBS
 - Add libasr include directory to CPPFLAGS
 - Add libasr lib directory to LDFLAGS

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2071,6 +2071,8 @@ fi
 
 
 CHECK_LIBASR([])
+LIBS="-lasr ${LIBS}"
+
 CHECK_LIBDB([])
 
 
--- a/m4/_check_libasr.m4
+++ b/m4/_check_libasr.m4
@@ -6,21 +6,35 @@ AC_DEFUN([WITH_LIBASR], [{
 	AC_ARG_WITH([libasr],
 		[  --with-libasr=PATH			Use libasr located in PATH],
 		[
-			with_libasr_include=${withval}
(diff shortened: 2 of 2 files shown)


poolpOrg:
nope:

1- I have reverted the configure.ac cleanups, there's a lot more to do and I can't break porable branch for too long, the work is done in branch "autohell"

2- I'm ok with what you're trying to do with the diff (support relative paths + various lib paths) but I'm trying to redesign these checks so they don't overwrite CPPFLAGS / LDFLAGS but set check-specific variables that configure.ac will use to set CPPFLAGS/LDFLAGS the way it wants:

    CHECK_LIBASR([])
    CHECK_LIBEVENT([])
    CHECK_LIBZ([])
    
    CPPFLAGS_smtpd="$with_libasr_cppflags $with_libevent_cppflags $with_libz_cppflags"
    CPPFLAGS_smtpctl="$with_libz_cppflags"
    CPPFLAGS_makemap="$with_libz_cppflags"

Right now, if you --with-libasr, smtpctl and makmap get linked against it, same applies to libdb, libz, etc ... this needs to end now because OpenSMTPD-extras is going to share the m4/ directory and you build filter-dnsbl and filter-mysql, you don't want both of them linked with both libs.

I opened a ticket #563 

I'll adapt your diff to branch autohell and rework it a bit to match the new model.
Thanks

PR #571, opened 2015-05-15
lko_session: Add format specifier for tag

kyrias:
Add a format specifier for getting the tag of the address, if any

--

Not all MDAs have support for tagged addresses built in, so this PR adds support for a format specifier that expands to the tag, if there is one.

### DIFF
--- a/smtpd/lka_session.c
+++ b/smtpd/lka_session.c
@@ -73,6 +73,7 @@ static void mailaddr_to_username(const struct mailaddr *, char *, size_t);
 static int mod_lowercase(char *, size_t);
 static int mod_uppercase(char *, size_t);
 static int mod_strip(char *, size_t);
+static int mod_tag(char *, size_t);
 
 struct modifiers {
 	char	*name;
@@ -81,6 +82,7 @@ struct modifiers {
 	{ "lowercase",	mod_lowercase },
 	{ "uppercase",	mod_uppercase },
 	{ "strip",	mod_strip },
+	{ "tag",	mod_tag },
 	{ "raw",	NULL },		/* special case, must stay last */
 };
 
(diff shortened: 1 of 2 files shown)


poolpOrg:
I'll test and merge tomorrow ;-)

poolpOrg:
Actually, I think your code is misplaced, it would make more sense if it the tag value was accessible as `%{sender.tag}`, `%{rcpt.tag}` and `%{dest.tag}`.

Any reason why you implemented it as a modifier ?

kyrias:
You know, you're totally right and I have no idea why I did it this way. Probably because it's the first thing I thought of and it was simple. I'll look the more sensible approach though.

Would you prefer adding it as another member of the `mailaddr` struct, or making it a separate function?

poolpOrg:
a separate function would be better  ;)

kyrias:
@poolpOrg https://github.com/kyrias/OpenSMTPD/commit/d8597bfa5d3e099f39 is a WIP version of `%{rcpt.tag}`. Any comments?

Currently it isn't what I'd call too great. For one, when there is no tag it has to return a space since it can't return an empty string, but I can't think of any other way to do it. Would also like comments on where to put the function and what to name it.

poolpOrg:
Have not forgotten about this one, just focusing on the release stress test, I'll get back to you as soon as I have some time to double check the logic

kyrias:
Of course, no rush. :)

poolpOrg:
The original PR doesn't bring the feature done right.

The comment at:

    https://github.com/OpenSMTPD/OpenSMTPD/pull/571#issuecomment-102665938

is a better way to do it but returning a 'space' when tag is empty is not acceptable either, it needs to be done differently.

I'll close the PR and if you still want this feature, please start a discussion on misc@ so we can discuss this publicly. I'm not too happy about adding a hack or half-working solution in the middle of this code path which is quite tricky already.

PR #590, opened 2015-06-12
Rename README to README.md

mulander:
The issue was introduced with this [commit](https://github.com/OpenSMTPD/OpenSMTPD/commit/667492876009201b4ab5cc824d28450412ad457b#diff-c47c7c7383225ab55ff591cb59c41e6b). I don't know if it was deliberate or an accident.

---
Github renders the README file based on it's extension. The OpenSMTPD README file is in markdown
format but is displayed as plain text which messes up both formatting and breaks some links on the page.

For example clicking the 'more information' URL results in a 404 from

http://www.opensmtpd.org/portable.html%5D(http://www.opensmtpd.org/portable.html)

instead of directing the user to

http://www.opensmtpd.org/portable.html

A simple file rename from README to README.md makes the formatting nicer and fixes the way Github
treats the links.


poolpOrg:
nope, we'll actually starting getting rid of markdown because this README ends up in tarballs and not everyone uses github.

mulander:
In that case I could go through it and remove the MD formatting.

Just let me know if you want someone to do the grunt work & if you have any preferences on the final layout :)

poolpOrg:
On Fri, Jun 12, 2015 at 07:02:41AM -0700, mulander wrote:
> In that case I could go through it and remove the MD formatting.
> 
> Just let me know if you want someone to do the grunt work & if you have any preferences on the final layout :)
> 

Please enjoy yourself ;-)


-- 
Gilles Chehade

https://www.poolp.org                                          @poolpOrg

mulander:
Closing as #591 removes the markdown formatting.

PR #604, opened 2015-06-27
Fix solaris m data

bsdsx:


poolpOrg:
This is not ok, 2 reasons:

First of all, we want the smtpd/ directory free of ifdef's so, unless for very specific cases which have a good rationale and for some headers, you have to solve this at the portable glue layer (either configure.ac or openbsd-compat/*).

Then, when you're forced to use conditional preprocessor macros, you must ensure that they do not test a system but that they test a feature. In this case, you're testing M_DATA and the sun/__sun are not really relevant, if someone forks your OS or comes with an OS that already has M_DATA, they will face the same issue.

So my suggestions:
1- check in configure.ac if M_DATA is defined, if it is then undef it from "includes.h"
2- drop the defined(sun) || defined(__sun)

PR #605, opened 2015-06-27
Fix compat imsg h

bsdsx:


bsdsx:
Reset to origin, merge upstream

PR #610, opened 2015-06-28
Fix path defpath

bsdsx:


poolpOrg:
I don't like this one either, IMO the path should be set in the sun specific configure.ac part and _PATH_DEFPATH should be set to that value without having system-specific ifdefs.

PR #613, opened 2015-07-10
[TLS] Use correct function to determine session protocol version

aaronmdjones:
SSL_get_cipher_version() returns the version of SSL/TLS that the
ciphersuite was introduced/defined in, not the version in use for
the connection.

Use SSL_get_version() instead.


aaronmdjones:
Also, why even includes bits=? It's not like that can't be inferred from the cipher suite name (at least, 99% of the time)

poolpOrg:
I'll merge your pull request shortly.

As for the bits part, it doesn't cost anything to display, in some cases where it's not obvious from the name (ie: ECDHE-RSA-CHACHA20-POLY130) it makes it clearer for the users.

PR #614, opened 2015-07-11
[TLS] Comply with RFC 7465 by default

aaronmdjones:
The manpage says that if the user does not configure any ciphers,
that ciphers "considered to be secure" will be used. However, RC4
is not secure, and RC4 is not forbidden by the default list.

This will still be needed even if LibreSSL removes (or has removed)
support for the RC4 cipher in TLS as this is the portable branch
and will likely be used with OpenSSL builds that haven't done so.


poolpOrg:
RC4 is not forbidden by the default list because... it is not part of the default list.
We require ciphers in the HIGH suite whereas RC4 is in MEDIUM:

both on LibreSSL:

    $ openssl version                                                                                                                                          
    LibreSSL 2.2
    $ openssl ciphers HIGH|grep RC4   
    $ openssl ciphers MEDIUM|grep RC4
    IDEA-CBC-SHA:ECDHE-RSA-RC4-SHA:ECDHE-ECDSA-RC4-SHA:AECDH-RC4-SHA:ADH-RC4-MD5:ECDH-RSA-RC4-SHA:ECDH-ECDSA-RC4-SHA:RC4-SHA:RC4-MD5
    $


and a relatively old OpenSSL on a Linux system:

    gilles:~$ openssl version
    OpenSSL 1.0.1k 8 Jan 2015
    gilles:~$ openssl ciphers HIGH|grep RC4
    gilles:~$ openssl ciphers MEDIUM|grep RC4
    DHE-RSA-SEED-SHA:DHE-DSS-SEED-SHA:ADH-SEED-SHA:SEED-SHA:ECDHE-RSA-RC4-SHA:ECDHE-ECDSA-RC4-SHA:AECDH-RC4-SHA:ADH-RC4-MD5:ECDH-RSA-RC4-SHA:ECDH-ECDSA-RC4-SHA:RC4-SHA:RC4-MD5:PSK-RC4-SHA
    gilles:~$ 

OpenSSL code from early 2014 shows it was already in MEDIUM back then.

aaronmdjones:
Oh, excellent. I did not know it had been moved to MEDIUM, even though the function of HIGH is "128 bits or higher", which RC4 is.

[Side note] Another interesting case, 3DES is in HIGH even though it only offers 112 bits (from a 168-bit key). It's all very crude in my opinion.

Thanks again!

PR #616, opened 2015-07-12
undef M_DATA if HAVE_M_DATA

bsdsx:


bsdsx:
Sorry, it's not the right fix.

bsdsx:
The last commit put undef in the right place.

PR #625, opened 2015-07-31
set length of smtp_session->username to SMTP_PATH_MAX

satan1st:
also check if the username is to long and log the failure. Now
a user can auth with any possible mailbox.

This fix the issue authing with username longen then 9 chars

### DIFF
--- a/smtpd/smtp_session.c
+++ b/smtpd/smtp_session.c
@@ -82,7 +82,7 @@ enum session_flags {
 	SF_BADINPUT		= 0x0080,
 	SF_FILTERCONN		= 0x0100,
 	SF_FILTERDATA		= 0x0200,
-	SF_USERTOLONG		= 0x0400,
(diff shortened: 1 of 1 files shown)


poolpOrg:
I bumped the limit to SMTPD_MAXMAILADDRSIZE but didn't go the SF_ way:

The whole point of the SF_USERTOOLONG flag is to try to hide that the username was rejected because it is too long. In doing so, you're passing a truncated buffer to the auth layer (which I'm opposed to) in order to prevent a timing-attack which would allow disclosing that the user was rejected because it was too long.

I think this is over-complicated for absolutely no benefit, a hacker that is smart enough to mount a timing-attack is certainly smart enough to look at the source and eliminate all usernames that are larger than SMTPD_MAX_MAILADDRSIZE.

We might as well reject them to start with, avoid the complexity and avoid passing truncated buffers down layers for nothing.

PR #634, opened 2015-08-27
Don't remove the BCC-field

satan1st:
According to the RCF 5322 there are some reasons a message
contains a Bcc-field.

This reverts commit 40aaa643bd035f132214e0e8a2d37f77a83a5af2.

According to the RFC 2 of 3 ways to implement Bcc will result in messages
with a Bcc-field.
see: https://tools.ietf.org/html/rfc5322#section-3.6.3

### DIFF
--- a/smtpd/smtp_session.c
+++ b/smtpd/smtp_session.c
@@ -152,7 +152,6 @@ struct smtp_session {
 	int			 rcvcount;
 	int			 dataeom;
 
-	int			 skiphdr;
 	struct event		 pause;
 
 	struct rfc2822_parser	 rfc2822_parser;
@@ -1412,18 +1411,6 @@ smtp_io(struct io *io, int evt)
 				line += 1;
 				len -= 1;
 			}
-
-			if (isspace((unsigned char)line[0]) && s->skiphdr)
-                                goto nextline;
-                        s->skiphdr = 0;
-
-                        /* BCC should be stripped from headers */
-                        if (! s->hdrdone) {
-                                if (strncasecmp("bcc:", line, 4) == 0) {
-                                        s->skiphdr = 1;
-                                        goto nextline;
-                                }
-                        }
 
 			log_trace(TRACE_SMTP, "<<< [MSG] %s", line);
 			smtp_filter_dataline(s, line);


poolpOrg:
Unfortunately it isn't that simple and reverting this commit is not really an option.
While the RFC dictates this behavior for clients, some clients have not followed it correctly and assume that the server will strip the BCC header. At least two MUA in common use by OpenBSD hackers expect this, but I suspect others do too because they copied each other.
As a result, not stripping BCC on the server and following strictly the RFC exposes users of such clients to leaking their BCC which is a far more annoying problem that the cosmetic appearance of a BCC header in two particular cases (case 2 & 3 are very uncommon).
This is one case where following the RFC strictly is hurting more people than it should
On Aug 28, 2015 12:47 AM, satan1st <notifications@github.com> wrote:According to the RCF 5322 there are some reasons a message
contains a Bcc-field.

This reverts commit 40aaa64.

According to the RFC 2 of 3 ways to implement Bcc will result in messages
with a Bcc-field.
see: https://tools.ietf.org/html/rfc5322#section-3.6.3



You can view, comment on, or merge this pull request online at:
  https://github.com/OpenSMTPD/OpenSMTPD/pull/634

Commit Summary
Don't remove the BCC-field

File Changes

    M
    smtpd/smtp_session.c
    (13)
  

Patch Links:
https://github.com/OpenSMTPD/OpenSMTPD/pull/634.patchhttps://github.com/OpenSMTPD/OpenSMTPD/pull/634.diff

satan1st:
[2015-08-27 16:55] Gilles Chehade <notifications@github.com>
> While the RFC dictates this behavior for clients, some clients have not
> followed it correctly and assume that the server will strip the BCC header.
> At least two MUA in common use by OpenBSD hackers expect this, but I suspect
> others do too because they copied each other.

Can you just say which MUAs you mean?

> As a result, not stripping BCC on the server and following strictly the RFC
> exposes users of such clients to leaking their BCC which is a far more
> annoying problem that the cosmetic appearance of a BCC header in two particular
> cases (case 2 & 3 are very uncommon).

Yes case 2 & 3 are very uncommon, but I don't think a MTA schould decide for all
users witch is the best option. Not even the RFC has done this decision.

IMHO the best way to fix this is to fix these old mailers.

They have two options: Use the sendmail-interface or cut out the Bcc field.

There is a way to fix this in OpenSMTPD, write a filter.

Philipp

poolpOrg:
> Can you just say which MUAs you mean ?

I only recall MH-E as it was the one I installed to confirm the issue was fixed, you should ask on the OpenBSD mailing list which MUA people use if you're interested in this. You can focus on the ones that do a direct network connection rather than use the sendmail interface, as these were the ones that hit the issue.


> Yes case 2 & 3 are very uncommon, but I don't think a MTA schould decide
> for all users witch is the best option. Not even the RFC has done this decision.

I don't agree.

We're not talking about a philosophical decision, we're facing a case where following the RFC strictly will cause a larger nuisance than not [0].

If we strip this header, like we do, most people don't notice. Those who notice only suffer a cosmetic glitch with no consequence in case 2 and 3.

If we don't strip this header, people using the wrong MUA will get bit and disclose their BCC without knowing it. They are going to realize that they disclosed their BCC at the same moment that their list of recipient also received the BCC list.

Following the RFC is good and we try to do it as much as possible, but interoperability, de-facto standards and ensuring we don't break existing software takes precedence. Users don't care about RFC, if their software used to work and it no longer works when they switch to OpenSMTPD because we follow the RFC too strictly, it means that we're being stricter than the others and we broke mail.


> IMHO the best way to fix this is to fix these old mailers.

Yes, in an ideal world, but that's not going to happen: I've used the same MUA for 15 years, others are using more ancient ones that are not necessarily still maintained, we're not going to change the MUA we use because some MTA has decided that we had to.

People want a working a MTA that is transparent to their users, not something that will require everyone to change habits.

Meanwhile we don't want people to get bit by BCC leakage.


> There is a way to fix this in OpenSMTPD, write a filter.

A filter for a default behavior that leaks sensitive data is really not the proper fix.


[0] there are other places where we diverge from RFC because it just is too much nuisance to everyone, the line length for headers and data part is a perfect example of how following the RFC too strictly because "we're better of fixing the MUA" just prevents mail from being accepted and causes mail loss.

satan1st:
[2015-08-29 06:01] Gilles Chehade <notifications@github.com>
> > Can you just say which MUAs you mean ?
> 
> I only recall MH-E as it was the one I installed to confirm the issue was fixed, you should ask on the OpenBSD mailing list which MUA peop
> le use if you're interested in this. You can focus on the ones that do a direct network connection rather than use the sendmail interface,
>  as these were the ones that hit the issue.

OK to be clear, MH-E is not a MUA, it's a front-end for MH-like MUAs. The default
on is nmh. I have checked the git-version of nmh and it don't contain this bug.
A tcp-dump is attached[0].

There are other MH-like MUAs, but I doupt any of them contain this
with intent.

> > Yes case 2 & 3 are very uncommon, but I don't think a MTA schould decide
> > for all users witch is the best option. Not even the RFC has done this decision.
> 
> I don't agree.
> 
> We're not talking about a philosophical decision, we're facing a case where following the RFC strictly will cause a larger nuisance than n
> ot [0].

I think this is an philosophical decision. I don't say follow without
thinking about it.  But you should be aware of other implementations
you mide break. There is a clause in the RFC[1], where this question
is discussed.

> If we strip this header, like we do, most people don't notice. Those who notice only suffer a cosmetic glitch with no consequence in case 
> 2 and 3.

If you read the ``Security Considerations''[1] of the RFC you will see
there are consequences.

> If we don't strip this header, people using the wrong MUA will get bit and disclose their BCC without knowing it. They are going to realiz
> e that they disclosed their BCC at the same moment that their list of recipient also received the BCC list.

If you strip out this header, people with the wrong MUA are facing a
feature remove without any reason.

> > IMHO the best way to fix this is to fix these old mailers.
> 
> Yes, in an ideal world, but that's not going to happen: I've used the same MUA for 15 years, others are using more ancient ones that are n
> ot necessarily still maintained, we're not going to change the MUA we use because some MTA has decided that we had to.

Of course not, but if a MTA doesn't work as expected you should consider to
fix the bug in your code, and not expect the MTA implement a workaround that
that breaks features of other MUAs.

> Meanwhile we don't want people to get bit by BCC leakage.

Of course this is bad, but it's not the task of the MTA to take care
of this.

> > There is a way to fix this in OpenSMTPD, write a filter.
> 
> A filter for a default behavior that leaks sensitive data is really not the proper fix.

I don't think this should be default behavior. What comes next
auto-encryption, fixing bad mime or spell checking? Some MUAs don't
implement this, so better fix it.

To make this clear, I don't think a MTA should change a mail in any
way by default, expect adding a Received-field or something similar.

Philipp

[O]: I hope github can handle this
[1]: https://tools.ietf.org/html/rfc5322#section-5 

220 freebsd-desktop.bureaucracy.de ESMTP OpenSMTPD
EHLO localhost
250-freebsd-desktop.bureaucracy.de Hello localhost [127.0.0.1], pleased to meet you
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-SIZE 36700160
250-DSN
250 HELP
MAIL FROM:<satanist@bureaucracy.de>
250 2.0.0: Ok
RCPT TO:<philipp@peter-tatacs.de>
250 2.1.5 Destination address valid: Recipient ok
RCPT TO:<philipp@bureaucracy.de>
250 2.1.5 Destination address valid: Recipient ok
RSET
250 2.0.0: Reset state
MAIL FROM:<satanist@bureaucracy.de>
250 2.0.0: Ok
RCPT TO:<philipp@peter-tatacs.de>
250 2.1.5 Destination address valid: Recipient ok
DATA
354 Enter mail, end with "." on a line by itself
From: "P. Takacs" <satanist@bureaucracy.de>
To: philipp@peter-tatacs.de
Subject: Test
Date: Sat, 29 Aug 2015 22:28:18 +0200

Das ist ein test
.
250 2.0.0: aa38bc2d Message accepted for delivery
RSET
250 2.0.0: Reset state
MAIL FROM:<satanist@bureaucracy.de>
250 2.0.0: Ok
RCPT TO:<philipp@bureaucracy.de>
250 2.1.5 Destination address valid: Recipient ok
DATA
354 Enter mail, end with "." on a line by itself
From: "P. Takacs" <satanist@bureaucracy.de>
Date: Sat, 29 Aug 2015 22:28:18 +0200
Subject: Test
BCC:

------- Blind-Carbon-Copy

From: "P. Takacs" <satanist@bureaucracy.de>
To: philipp@peter-tatacs.de
Subject: Test
Date: Sat, 29 Aug 2015 22:28:18 +0200

Das ist ein test

------- End of Blind-Carbon-Copy
.
250 2.0.0: 939b47f5 Message accepted for delivery
QUIT
221 2.0.0: Bye

poolpOrg:
> OK to be clear, MH-E is not a MUA, it's a front-end for MH-like MUAs. The default on is nmh. I have checked the git-version of nmh and it don't contain this bug.

Yet 3/4 different OpenBSD hackers (theo, tedu, otto, millert) have complained about this bug and confirmed it gone, should we assume they are lying about experiencing it ?


> I think this is an philosophical decision. I don't say follow without thinking about it.  But you should be aware of other implementations you mide break. There is a clause in the RFC[1], where this question is discussed.
> [...]
> If you read the ``Security Considerations''[1] of the RFC you will see there are consequences.

I love how you imply we don't pour thinking into decisions we make for OpenSMTPD, let me put a bit of perspective:

Your decision to revert this commit is based on a paragraph from the RFC, your views of what the MTA should do and your thinking that it is a philosophical decision.

Our decision to commit this is based on user complaints, about 20 exchanges with 4 developers followed by a mail to all openbsd hackers to describe what was about to be committed, a discussion because not everyone agreed (and as might surprise you, I was initially not convinced we should do it...), a look at what that would imply RFC-wise, a comparison of behavior with Postfix and Sendmail, a test of several MUA to see if they would be impacted and finally snapshots so a wider audience would test and report issues.

We don't just put code without thinking and discussions, this is not how we work.


> If you strip out this header, people with the wrong MUA are facing a feature remove without any reason.

This is called a trade-off, yes, do we prefer people leaking their BCC vs people not seeing a cosmetic header. Not stripping it leads to people getting bit by leakage, stripping it led to absolutely no complaint whatsoever. Which one has a higher security consideration in your opinion ?

The security consideration in the RFC is bullshit, ever since BCC was invented people have replied to all in disregard of whether or not they were part of the advertised recipients. Do you think protecting against this, which will still happen no matter what we do (and which has never generated a single complaint), is worth deciding that we should allow BCC leakage if people use the wrong MUA ?

We don't think so and those who complained neither.


> I don't think this should be default behavior. What comes next auto-encryption, fixing bad mime or spell checking? Some MUAs don't implement this, so better fix it.

This is not the same: BCC is a bastard feature which mixes DATA and protocol, so comparing it to a DATA-only feature when speaking in the scope of a daemon implementing the protocol is seeing half the problem.

But let me tell you a couple things:

We already fix bad mime, we assume 8-bit bytes no matter if the client has requested 8BITMIME or not. If we don't do this, some clients break. We could argue that they should be fixed, but it turns out pretty much the entire world is 8-bit bytes now and not doing so because we want a strict RFC compliance is going to annoy people whereas no one will ever notice if we are 7-bit strict.

We also do append-domain by default, allowing MAIL FROM:<gilles> to be automagically expanded to MAIL FROM:<gilles@poolp.org>. If we don't, some clients break and interop with other MTA leads to strange headers. We shouldn't do this, they should be fixed, but it turns out that this breaks mail to users.


> To make this clear, I don't think a MTA should change a mail in any way by default, expect adding a Received-field or something similar.

Yes, unfortunately that's not how mail works, if you don't change a mail, you break exchanges. People are using mail because it works, not because it follows a RFC strictly.

satan1st:
[2015-08-31 00:37] Gilles Chehade <notifications@github.com>
> > OK to be clear, MH-E is not a MUA, it's a front-end for MH-like MUAs. The default on is nmh. I have checked the git-version of nmh and i
> t don't contain this bug.
> 
> Yet 3/4 different OpenBSD hackers (theo, tedu, otto, millert) have complained about this bug and confirmed it gone, should we assume they 
> are lying about experiencing it ?

Quote from MH-19910201.txt[0]
> and  the  "Bcc:"  field will be removed from the message sent to
> sighted recipients.

What shall I assume, that these people are to lazy too read the documentation
of there Software?

I don't believe you, if you say there is a MH-mailer which assume this
behavior. Probably there are some which contain a bug and leak this
headers, depending on the config.

I found this lines in MH-6.8.5
> if (!fill_in && (flags & (HBCC | HDCC)) || mp -> m_ingrp)
> return 1;

Probably this '!fill_in &&' can be removed, because it's from
an old and removed feature.

> > I think this is an philosophical decision. I don't say follow without thinking about it.  But you should be aware of other implementatio
> ns you mide break. There is a clause in the RFC[1], where this question is discussed.
> > [...]
> > If you read the ``Security Considerations''[1] of the RFC you will see there are consequences.
> 
> I love how you imply we don't pour thinking into decisions we make for OpenSMTPD, let me put a bit of perspective:

I don't imply this, I anwser your wrong claims. You said this is not
philosophical, I explained why I think it is. You said there is no
impact, I refereed you were I see an impact.

> We don't just put code without thinking and discussions, this is not how we work.

Now you imply I have now idea of Software Development. Don't do this.

> > If you strip out this header, people with the wrong MUA are facing a feature remove without any reason.
> 
> This is called a trade-off, yes, do we prefer people leaking their BCC vs people not seeing a cosmetic header. Not stripping it leads to p
> eople getting bit by leakage, stripping it led to absolutely no complaint whatsoever. Which one has a higher security consideration in you
> r opinion ?
> 
> The security consideration in the RFC is bullshit, ever since BCC was invented people have replied to all in disregard of whether or not t
> hey were part of the advertised recipients. Do you think protecting against this, which will still happen no matter what we do (and which 
> has never generated a single complaint), is worth deciding that we should allow BCC leakage if people use the wrong MUA ?
> 
> We don't think so and those who complained neither.

OK, a more technical point:
This code can cause DoS. If a DKIM-signed mail has signed the BCC-field,
a mailserver can drop this mail or blacklist the hole sever.

Furthermore, you forgot half the problem. There is a Resend-Bcc-field.
Also there is the Dcc implementation of MH, these headers are probably
leaked, too.

What about forwarded mails, containing a Bcc-field, attached mboxes, ...?
As you said, you can't prevent against a dump User/Client.

This is the kind of smart software I don't like, software which thinks
it knows better what I want.

With a filter you only have some of these problems. You can
choose on witch interface the the field should be stripped,
so there is no DoS. If you have forgot a field you can extend
the filter.

> > I don't think this should be default behavior. What comes next auto-encryption, fixing bad mime or spell checking? Some MUAs don't imple
> ment this, so better fix it.
> 
> This is not the same: BCC is a bastard feature which mixes DATA and protocol, so comparing it to a DATA-only feature when speaking in the 
> scope of a daemon implementing the protocol is seeing half the problem.
> 
> But let me tell you a couple things:
> 
> We already fix bad mime, we assume 8-bit bytes no matter if the client has requested 8BITMIME or not. If we don't do this, some clients br
> eak. We could argue that they should be fixed, but it turns out pretty much the entire world is 8-bit bytes now and not doing so because w
> e want a strict RFC compliance is going to annoy people whereas no one will ever notice if we are 7-bit strict.
> 
> We also do append-domain by default, allowing MAIL FROM:<gilles> to be automagically expanded to MAIL FROM:<gilles@poolp.org>. If we don't
> , some clients break and interop with other MTA leads to strange headers. We shouldn't do this, they should be fixed, but it turns out tha
> t this breaks mail to users.

This is not the same, these fixes doesn't change the behavior for a
client which act according to the RFC.

With this point I meant something like re encode everything
to base64, like the Microsoft Exchange version

> > To make this clear, I don't think a MTA should change a mail in any way by default, expect adding a Received-field or something similar.
> 
> Yes, unfortunately that's not how mail works, if you don't change a mail, you break exchanges. People are using mail because it works, not
>  because it follows a RFC strictly.

Then let me ask you a question: Why do we need a standard, if
everyone writes a workaround for not standard implementations?

[0]: http://git.savannah.gnu.org/cgit/nmh.git/tree/docs/historical/MH-19910201.txt

poolpOrg:
> Now you imply I have now idea of Software Development. Don't do this.

no, I'm not implying anything.

you saw something you didn't like and instead of asking why we made that choice and what the rationale was, you provided a diff to remove it plain and simple with no rationale but a link to the RFC.

when I told you people had complained about being bit by this, you ruled it out because you think it's wrong and pointed us to the RFC again telling us we should read this and that, and think about this and that, still without asking about the rationale.

All I did was pointing out that our decision was not "just a commit" and that reverting it is not going to happen with just a commit either.

I will say it again:

your rationale stands, and *I dislike stripping BCC*, but whether you like it or not, or think the users are lazy, or don't believe them or me that they hit this issue, they did. you can go dig whatever git commit and whatever README, it won't change this fact no matter how much you'd like to prove me wrong.

what you are asking with this pull request is that we ignore their problem on the grounds that the RFC says otherwise and that they must fix their client. however, what you don't seem to grasp is that THIS yield to several complaints whereas stripping never did, and that we work to provide a solution for the larger bunch, not to make one person happy.

I would be happy with removing the stripping, just not at the cost of dealing with more complaints because more people get annoyed by the lack of stripping than not.

Note that I didn't reject your pull request and closed it with a WONTFIX, I told you that you could take it publically and confront the people who are experiencing these BCC leaks. If you manage to convince them and others that we should let them leak, then I'd be more than willing to apply. Until then, as far as I'm concerned, there are more people annoyed with the lack of stripping than the opposite.


> Then let me ask you a question: Why do we need a standard, if everyone writes a workaround for not standard implementations?

Because not every MTA/MUA follows the standard, yet every user wants its mails to pass through and we're writing a software to accommodate the users not the IETF.

If we followed the RFC too strictly, we'd lose mail that currently turns out to exchange just fine with all the MTAs I know.

poolpOrg:
This won't be merged unless a discussion takes place on misc@openbsd.org and theo & others are convinced that we should let MUA who don't strip bcc disclose their bcc list.

I'm closing the pull request since it's already conflicting with current code, if the discussion takes place and decision is taken to apply this, I'll adapt and credit you.

PR #636, opened 2015-10-03
fix getc() usage in compat fgetln()

timotejl:
If return value from getc() is stored in a char, comparison to EOF never
succeeds on some platforms.


poolpOrg:
we had discussed this yesterday and planned to push a fix but you beat us to it.

Thanks ;-)

PR #649, opened 2016-01-11
Add warning feature for table permission at runtime

gunhu:


poolpOrg:
was actually okayed jung@ already

PR #651, opened 2016-02-05
Solaris sun identifier

bsdsx:
Don't use 'sun' as identifier


bsdsx:
bad pull request. close it.

PR #654, opened 2016-02-06
Don't use 'sun' as identifier

bsdsx:
On Solaris, don't use 'sun' as identifier


poolpOrg:
nevermind, I made the adaptation and credited you with the commit

PR #656, opened 2016-02-13
check for WAIT_MYPGRP

bsdsx:
Solaris 10 don't define WAIT_MYPGRP


poolpOrg:
two comments:

1- does setting WAIT_MYPGRP to 0 achieve the same on Solaris ?

2- can you add a comment above this check + the AF_LOCAL one to make it clear this was introduced for Solaris 10 ?

poolpOrg:
verified.

PR #661, opened 2016-02-27
u_int16_t, u_int32_t

bsdsx:
u_int16_t and u_int32_t  are not define on solaris


poolpOrg:
actually the real question is ... what makes use of these instead of uint16_t and uint32_t

we should fix these instead

poolpOrg:
these were only used by imsg.h/imsg.c, I fixed the files so they use the proper types.

PR #666, opened 2016-03-12
solaris don't have tm_gmtoff/tm_zone

bsdsx:


poolpOrg:
I don't like this diff, to.c becomes unreadable, can we work out a different solution like possibly functions in utils.c which are then conditionally used in to.c ?

poolpOrg:
err, there is a way by simply rearranging your code so that you don't need the whole ifdef dance without moving code to utils.c:

- simply use a temporary variable and assign either one of the struct member or your long variable
- then use that temporary variable in the parameters rather than ifdef-ing large chunks.

bsdsx:
Hello,

if my code match your last comment, could you merge this pr ?
i can explain it if needed

poolpOrg:
I'll merge today, I just need to do a bit of rework in master branch first to reduce the delta with your portable change.

poolpOrg:
done

PR #670, opened 2016-03-20
set IPV6_V6ONLY if available

valeriangalliat:
When trying to listen for exanple on both `0.0.0.0` and `::`, the IPv6 listen would fail with "Address already in use" because it's also trying to bind on IPv4 for mapping IPv4 to IPv6.

I'm really not sure how to handle this, but it seems like the other softwares I'm running (Dovecot, NGINX) all explicitly set `IPV6_V6ONLY` to `1` (see for [Dovecot](https://github.com/dovecot/core/blob/7c925149e49f7cce41c90d562ff3835b66ddca29/src/lib/net.c#L445) and [NGINX](https://github.com/nginx/nginx/blob/2ce791f2cddff967fd3bcbbedcd4ea283a9c77e1/src/core/ngx_connection.c#L490)).

This PR adds this option for IPv6 sockets if supported.

### DIFF
--- a/smtpd/smtp.c
+++ b/smtpd/smtp.c
@@ -160,6 +160,15 @@ smtp_setup_listeners(void)
 			sizeof(opt)) < 0)
 			fatal("smtpd: setsockopt");
 #endif
+#ifdef IPV6_V6ONLY
+		/* If using IPv6, bind only to IPv6 if possible. This avoids
+		   ambiguities with IPv4-mapped IPv6 addresses. */
+		if (l->ss.ss_family == AF_INET6) {
+				if (setsockopt(l->fd, IPPROTO_IPV6, IPV6_V6ONLY, &opt,
+						sizeof(opt)) < 0)
+						fatal("smtpd: setsockopt");
+		}
+#endif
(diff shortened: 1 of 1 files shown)


poolpOrg:
Thanks

Diff reads ok to me but I'd like it reviewed by a regular IPv6 hacker to make sure there's no other way that would avoid adding a ifdef in the portable version of OpenSMTPD.

There's a hackathon next week so if I don't get a reply by then, it'll be merged during hackathon.

valeriangalliat:
Thanks @poolpOrg! In the end could you have the code reviewed by an IPv6 hacker?

poolpOrg:
kind of.

I reviewed the impact of this diff and it cannot break something that wasn't already broken:

- if inet4, nothing changes
- if inet6, nothing changes
- if ipv6_v6only, then it essentially solves something currently broken.

I double checked with a fellow hacker as to the impact on OpenBSD, and it essentially will change nothing for us.

valeriangalliat:
Awesome, thanks a lot for reviewing!

PR #688, opened 2016-05-02
fix openssl version check done wrong with an invalid mask

rosorio:
opensmtp checks the openssl version at startup,
but because the check is done wrong using an bad
mask opensmtp doesn't start.

This issue affects FreeBSD 10


poolpOrg:
I'm currently preparing release, I'll test this a bit before merging because it's already been fixed once and I need to make sure I don't see a regression elsewhere.

Thanks for the diff.

poolpOrg:
we'll test this in upcoming snapshots, thanks

PR #697, opened 2016-05-23
Add ability to set PATH_LIBEXEC at runtime via env var

obadz:
Originally written by @gebner for NixOS

The location where filters & co are going to be can't always be known at compile time. This patch adds the ability to override `PATH_LIBEXEC` at runtime using the `OPENSMTPD_PROC_PATH` environment variable.

### DIFF
--- a/smtpd/parse.y
+++ b/smtpd/parse.y
@@ -2534,13 +2534,19 @@ create_filter_proc(char *name, char *prog)
 {
 	struct filter_conf	*f;
 	char			*path;
+	const char		*proc_path;
 
 	if (dict_get(&conf->sc_filters, name)) {
 		yyerror("filter \"%s\" already defined", name);
 		return (NULL);
 	}
 
-	if (asprintf(&path, "%s/filter-%s", PATH_LIBEXEC, prog) == -1) {
+	proc_path = getenv("OPENSMTPD_PROC_PATH");
+	if (proc_path == NULL) {
+		proc_path = PATH_LIBEXEC;
+	}
+
+	if (asprintf(&path, "%s/filter-%s", proc_path, prog) == -1) {
 		yyerror("filter \"%s\" asprintf failed", name);
 		return (0);
 	}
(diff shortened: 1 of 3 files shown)


poolpOrg:
I'm sorry but I'm not convinced by neither one of the use-case and diff.

Setting the use-case aside for now, we will not accept diffs that start introducing changes of behavior based on environment variables, this is not how the rest of the code works and the overrides are only acceptable when done in configuration file or through getopt() flags.

As for the use-case, I don't see why they "can't always be known at compile time", this is a daemon dependency which is not so different from sysconfdir, PATH_SMTPCTL or even PATH_SOCKET. Why can they be known at compile time and not PATH_LIBEXEC ?

We're talking about a directory that contains programs that are going to be executed by the only privileged process in smtpd, I don't feel comfortable at all making this tweakable at runtime, this feels like a recipe for a disaster.

gebner:
For NixOS, a getopt() flag in smtpd would be fine as well.  We just need a way to tell smtpd where to look filters without recompiling it.

The reason we have added this environment variable is because on NixOS, there is no global /usr/libexec directory.  The directory containing the filters is generated based on the user configuration (i.e. which extras package is to be included, and with what configuration flags); the name of the directory is a hash of the configuration.  We then pass the location of the generated filter directory and generated configuration to the smtpd daemon:
```
      serviceConfig.ExecStart = "${opensmtpd}/sbin/smtpd -d -f ${conf} ${args}";
      environment.OPENSMTPD_PROC_PATH = "${procEnv}/libexec/opensmtpd";
```
(${conf} and ${procEnv} are dynamically generated based on the configuration)

> sysconfdir, PATH_SMTPCTL or even PATH_SOCKET. Why can they be known at compile time and not PATH_LIBEXEC

The configuration file is already overridable (and we make use of that).  We know where smtpctl will end up because it is in the same package as smtpd (and doesn't depend on the user configuration). PATH_SOCKET is fine, because we have a globally writable /run directory.  However the filters are a) in a different package and b) depend on user configuration, hence we do not know where they will end up when we compile opensmtpd.

Hope this clarifies the situation a bit.

poolpOrg:
Yes, i'm still a bit perplexed but not opposed to reviewing and discussing PR that implement such a feature as a getopt() option.

If you work on such a feature, look at how we did for plugging queue backends (one option, argument is subsystem=/path) and make sure to do it for filters, queues, tables, schedulers, not just for one subset.

The PR as is today is not acceptable so I'll close it and wait for a new one if you work on that feature :-)

PR #699, opened 2016-05-25
Support multiline responses from lmtp servers.

### LINKED ISSUE #694: LMTP broken: "illegal seek"
## Env
- Worked in: `5.7.3p1`
- Broken since: `5.9.2p1`
- System: Debian 8.4

## Description
LMTP delivery to local unix socket is broken. [This `fflush` call](https://github.com/OpenSMTPD/OpenSMTPD/blame/8b849a926e4bc7c6a6c6877b5930f7acc67c2aeb/smtpd/delivery_lmtp.c#L187) ends up with `Illegal seek`, same holds for the `getline` call afterwards (in case you comment out the `fflush`). It seems that LMTP stream under the former version were handled differently.

parchd-1:
Fixes #694 in which multiline responses from dovecot (and probably other LMTP servers) were not being anticipated by OpenSMTPD.

With this little change, OpenSMTPD will continue to read lines until the 4th character (i.e. character after the status code) is a space, indicating no more lines will be added to the response.

(first github pull request, hopefully I've done this right!)

### DIFF
--- a/smtpd/delivery_lmtp.c
+++ b/smtpd/delivery_lmtp.c
@@ -187,10 +187,10 @@ lmtp_cmd(char **buf, size_t *sz, int code, FILE *fp, const char *fmt, ...)
 	if (fflush(fp) != 0)
 		err(1, "fflush");
 
-    do {
-	if ((len = getline(buf, sz, fp)) == -1)
-		err(1, "getline");
-    } while((*buf)[3] != ' ');
+	do {
+		if ((len = getline(buf, sz, fp)) == -1)
+			err(1, "getline");
+	} while((*buf)[3] != ' ');
 
 	bufp = *buf;
 	if (len >= 2 && bufp[len - 2] == '\r')
--- a/smtpd/delivery_lmtp.c
+++ b/smtpd/delivery_lmtp.c
@@ -36,6 +36,8 @@
 
 #include "smtpd.h"
 
+#define MAX_CONTINUATIONS 20
+
 static int	inet_socket(char *);
 static int	lmtp_cmd(char **buf, size_t *, int, FILE *, const char *, ...)
 		    __attribute__((__format__ (printf, 5, 6)))
@@ -175,6 +177,7 @@ lmtp_cmd(char **buf, size_t *sz, int code, FILE *fp, const char *fmt, ...)
 	va_list	 ap;
 	char	*bufp;
 	ssize_t	 len;
+	unsigned int continuations = 0;
 
 	va_start(ap, fmt);
 	if (vfprintf(fp, fmt, ap) < 0)
@@ -192,6 +195,8 @@ lmtp_cmd(char **buf, size_t *sz, int code, FILE *fp, const char *fmt, ...)
 			err(1, "getline");
 		if (len < 4)
 			errx(1, "Server response line too short");
+		if (continuations++ >= MAX_CONTINUATIONS)
+			errx(1, "Too many continuations in response");
 	} while((*buf)[3] != ' ');
 
 	bufp = *buf;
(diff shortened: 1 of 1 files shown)


poolpOrg:
I need to double check as I'm reading on my phone, but it looks like diff introduces a bug possibly leading to the crash of mda. 
Gilles 
Le 25 mai 2016 11:25 AM, parchd-1 <notifications@github.com> a écrit :Fixes #694 in which multiline responses from dovecot (and probably other LMTP servers) were not being anticipated by OpenSMTPD.

With this little change, OpenSMTPD will continue to read lines until the 4th character (i.e. character after the status code) is a space, indicating no more lines will be added to the response.

(first github pull request, hopefully I've done this right!)



You can view, comment on, or merge this pull request online at:
  https://github.com/OpenSMTPD/OpenSMTPD/pull/699

Commit Summary
Support multiline responses from lmtp servers.

File Changes

    M
    smtpd/delivery_lmtp.c
    (2)
  

Patch Links:
https://github.com/OpenSMTPD/OpenSMTPD/pull/699.patchhttps://github.com/OpenSMTPD/OpenSMTPD/pull/699.diff

parchd-1:
I didn't think so, but maybe the bug you can see relates to error handling? I've looked again and realised there could be a problem depending on how OpenSMTPD deals with errors - something I haven't got time to check right now.

Or was it something else?

poolpOrg:
confirmed, there are two things wrong with the diff:

1- you access the 4th byte from line without checking that line returned  by getline() is long enough. if the LTMP server replies with a line shorter than 4 bytes, then you provoke an out-of-bounds access which will result in something  bad.

2- indentation seems weird, maybe it's github, but make sure indentation (spaces/tabs) match the coding style.

parchd-1:
1. Oh, you're right of course!
I'll sort that out.
2. I'll change the spaces for tabs - I didn't notice they didn't match.

parchd-1:
Hows that?
Now raises an error if the response is less than length 4. Anything less than that is an invalid response anyway because the response code is 3 characters already.

poolpOrg:
yes, that is much better, i'll get an ok from @snimmagadda and Joerg Jung and merge this.

poolpOrg:
humpf, I did another read in context and diff is still incorrect:

- buf is a (char **) not a (char *) therefore the check at offset 3 is incorrect
- as a side note, the check should be more strict

I'll come up with a variation of your diff and commit today

poolpOrg:
Second thought before we can merge:

You need to explicitely check for '-' to determine that it is a continuation line, right now your code says that I can send 250!foobar to continue. Also, you need to introduce a limit so that you don't allow a LMTP server to provide an infinite stream of continuation lines.

Do it as two commits, one that basically does:

    do {
        getline()
        if c == '-': continue
        else if c == ' ': break
        else error out
    } while (1);

And another one that adds a MAX_CONTINUATIONS define and which checks it inside the while loop, this way it not only fixes your issue buts makes it more robust.

parchd-1:
I did wonder whether the check should be for the hyphen or the space. One of the many RFCs I looked at for either SMTP or LMTP (I've lost track of which one it was in) mentioned just looking for a response code followed by a space. I've since seen in RFC 5321 that we should be prepared to accept the reply code without a space because broken clients might send just, e.g. "250\r\n" without the text that the protocol requires... what do you think about continuing for "-" and breaking for anything else?

Is it really right to add a MAX_CONTINUATIONS? What would the maximum be?

poolpOrg:
We're going to commit a fix today, it's a bit different from your diff as I spotted another issue but it's greatly inspired by this PR so I'll credit you ;-)

I'll close PR and ticket #694 when it's committed.

poolpOrg:
committed and synced to github, next snapshot and release will have this available.

PR #706, opened 2016-06-27
Flock

bsdsx:
Solaris don't have flock function


poolpOrg:
not merging 1907 commits, try again !

PR #708, opened 2016-07-02
add defines if not have flock

bsdsx:
add some define if not have flock


bsdsx:
add flock declaration
fix style

PR #719, opened 2016-09-17
Fix expression

bsdsx:
As explain in
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Generic-Declarations.html#Generic-Declarations

"Unlike the other `AC_CHECK_*S' macros, when a symbol is not
declared, HAVE_DECL_symbol is defined to `0' instead of
leaving HAVE_DECL_symbol undeclared."

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -346,7 +346,7 @@ AC_CHECK_DECLS([O_NONBLOCK], [], [], [
 #endif
 ])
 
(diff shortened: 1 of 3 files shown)


bsdsx:
add check for PF_LOCAL

poolpOrg:
merged, thanks !

PR #726, opened 2016-09-24
fix format long int

bsdsx:


poolpOrg:
I still have hope we get rid of testing for specific openssl versions though...

PR #731, opened 2016-10-08
Fix tabs/spaces, build function

bsdsx:
Fix tabs/spaces
Don't build function in compat if system support it


poolpOrg:
I'll merge this and publish a snapshot tomorrow to make sure we don't introduce a regression

PR #756, opened 2016-12-30
Remove errc

bsdsx:
errc is used only twice. Replace with errx.
openbsd-compat/errc.c is *NOT* removed by this PR


poolpOrg:
Unless there's a good reason not to use errc(), I'd prefer that we use it as it is the case in OpenBSD and we don't want to create unnecessary deltas between native and portable.

PR #763, opened 2017-01-11
fix compatibility with libressl

### LINKED ISSUE #691: configure script detects LibreSSL as OpenBSD stdlib
Since LibreSSL defines functions `reallocarray` etc, **but does not provide prototypes**, the current OpenSMTPD configure assumes it can use these and falls back to the implicit prototypes, which results in a pointer trunction for `reallocarray` and instant segfault in `sendmail(1)` when sending a mail (from `rcpt_add`).

I think the proper way to fix this is to use `AC_CHECK_DECLS` instead of `AC_CHECK_FUNCS` to ensure prototypes are there for all relevant functions.

smaeul:
These functions are exported by libcrypto from libressl, due to its
similar OpenBSD compatibility layer, but they are not present in any
header files. Thus, while we can use the existing compiled function,
and do not need to provide our own, we do need to provide the prototype
for it.

This avoids implicit function declarations and the resulting crashes due
to pointer truncation.

The patch is based on an equivalent patch for OpenSSH from
https://bugzilla.mindrot.org/show_bug.cgi?id=2465
Also see
https://github.com/libressl-portable/portable/issues/109

Fixes #691

### DIFF
--- a/openbsd-compat/defines.h
+++ b/openbsd-compat/defines.h
@@ -422,15 +422,6 @@ typedef uint16_t	in_port_t;
 #define INET6_ADDRSTRLEN 46
 #endif
 
-/*
- * Platforms that have arc4random_uniform() and not arc4random_stir()
- * shouldn't need the latter.
- */
-#if defined(HAVE_ARC4RANDOM) && defined(HAVE_ARC4RANDOM_UNIFORM) && \
-    !defined(HAVE_ARC4RANDOM_STIR)
-# define arc4random_stir()
-#endif
-
 #ifndef HAVE_VA_COPY
 # ifdef HAVE___VA_COPY
 #  define va_copy(dest, src) __va_copy(dest, src)
--- a/openbsd-compat/openbsd-compat.h
+++ b/openbsd-compat/openbsd-compat.h
@@ -119,20 +119,25 @@ int	BSDoptind;		/* index into parent argv vector */
 int getpeereid(int , uid_t *, gid_t *);
 #endif 
 
-#ifdef HAVE_ARC4RANDOM
-# ifndef HAVE_ARC4RANDOM_STIR
-#  define arc4random_stir()
-# endif
-#else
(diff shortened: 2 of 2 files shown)


poolpOrg:
indeed, merged, thanks !

4a6f656c:
I would not recommend doing this - libressl now has explicit symbol lists (for example, https://github.com/libressl-portable/openbsd/blob/master/src/lib/libcrypto/Symbols.list) and will only export public symbols (@busterb, @kinichiro - it seems that crypto/Makefile.am may be missing the glue to enable it, plus some additional symbols appear to be added for Windows, but in a file that is likely to be used on multiple platforms).

This is not yet in libressl-portable stable, but will be around May 2017, at which point this diff will likely create problems rather than fix them... I'm not sure of a good solution here, unless you condition on LIBRESSL_VERSION.

poolpOrg:
On Thu, Jan 12, 2017 at 04:50:51AM -0800, Joel Sing wrote:
> I would not recommend doing this - libressl now has explicit symbol lists (for example, https://github.com/libressl-portable/openbsd/blob/master/src/lib/libcrypto/Symbols.list) and will only export public symbols (@busterb, @kinichiro - it seems that crypto/Makefile.am may be missing the glue to enable it, plus some additional symbols appear to be added for Windows, but in a file that is likely to be used on multiple platforms).
> 
> This is not yet in libressl-portable stable, but will be around May 2017, at which point this diff will likely create problems rather than fix them... I'm not sure of a good solution here, unless you condition on LIBRESSL_VERSION.
> 

I suppose libressl-portable stable will be released after OpenBSD no ?

We're going to release opensmtpd -portable in May, after the OpenBSD
release takes place, so we could do the following:

1- we retain this "temporary" fix so our OpenSMTPD -portable snapshots
   benefit from the quick fix

2- we postpone our release so it happens after libressl -portable, so
   we can adapt to it

How does it sound to you ?

-- 
Gilles Chehade

https://www.poolp.org                                          @poolpOrg

PR #764, opened 2017-01-11
fix musl compatibility (missing function prototypes)

### LINKED ISSUE #758: musl-libc compatibility
**inet_net_pton**
musl-libc doesn't provide the inet_net_pton function. This is not an issue since it's in openbsd-compat/. However, in openbsd-compat.h it's missing, resulting in the following warning:

> ../../smtpd/to.c: In function 'text_to_netaddr':
> ../../smtpd/to.c:293:9: warning: implicit declaration of function 'inet_net_pton' [-Wimplicit-function-declaration]
>   bits = inet_net_pton(AF_INET, s, &ssin.sin_addr,
> 

Something like 

```
#ifndef HAVE_INET_NET_PTON
int inet_net_pton(int af, const char *src, void *dst, size_t size);
#endif 

```
should be added to openbsd-compat/openbsd-compat.h

****
**res_hnok**

> ../../smtpd/util.c: In function 'valid_domainpart':
> ../../smtpd/util.c:527:9: warning: implicit declaration of function 'res_hnok' [-Wimplicit-function-declaration]
>   return res_hnok(s);
>          ^~~~~~~~

It seems res_hnok is also not provided by musl, so it probably should be added to openbsd-compat.

smaeul:
inet_net_pton is already compiled, but no prototype is provided.
res_hnok is provided by the compatibility layer in libasr.

These fixes avoid warnings about implicit function declaration.

Fixes #758

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -594,6 +594,7 @@ AC_CHECK_FUNCS([ \
 	pledge \
 	pw_dup \
 	reallocarray \
+	res_hnok \
 	setenv \
 	setlinebuf \
 	setproctitle \
--- a/openbsd-compat/openbsd-compat.h
+++ b/openbsd-compat/openbsd-compat.h
@@ -208,8 +208,16 @@ void *reallocarray(void *, size_t, size_t);
 void errc(int, int, const char *, ...);
 #endif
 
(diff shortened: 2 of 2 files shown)


poolpOrg:
merged thanks !

PR #776, opened 2017-03-21
fix __progname detection/use

bsdsx:
Fix warning when check for __progname in configure.ac

Fix segfault when using undefined var in smtpctl.c


poolpOrg:
ssh_get_progname() ?

we should certainly rename it to smtpd_ or portable_, I'll merge your pull request but it would be nice if you could also submit a PR to change the name

Issue #709, opened 2016-07-03
Feature request: Support `senders` and `received-auth` in `listen on socket`


Wolf480pl:
Currently `listen on socket` accepts only `filters` option.
It would be useful if it also accepted `senders` and `received-auth` options, which would behave like in regular `listen` directives.

poolpOrg:
I'll look into this next week during the hackathon ;-)

Wolf480pl:
I started looking through the source code to see if it's been already implemented, and if not, how hard it would be to implement.

Apparently, if the config grammar was slightly changed to allow these two options in `listen on socket`, they would almost work, except `s->username` on the `smtp_session` would be empty.

Looks like the control process [already checks][0] the user's effective uid with `getpeereid`, and [sends it][1] to the pony process, so it's just a matter of taking it from [here][2], resolving it to a username and putting the username in the `smtp_session` somewhere around [here][3]?

Oh, and the uid->username lookup would have to go through the lookup process, cause it requires `/etc/passwd` access...

[0]: https://github.com/OpenSMTPD/OpenSMTPD/blob/portable/smtpd/control.c#L341
[1]: https://github.com/OpenSMTPD/OpenSMTPD/blob/portable/smtpd/control.c#L483
[2]: https://github.com/OpenSMTPD/OpenSMTPD/blob/portable/smtpd/smtp.c#L90
[3]: https://github.com/OpenSMTPD/OpenSMTPD/blob/portable/smtpd/smtp_session.c#L664

ngortheone:
@poolpOrg this looks a bit old. 
What is the current status and are there any future plans regarding this?

poolpOrg:
The PR was not adding the feature correctly and adding it correctly is tricky.

The status is "this might require a rework of how we consider the socket in the daemon to be done right", so not closing as wontfix but keeping open as don't hold your breath.

katp32:
Any update on this? Being able to use `senders` in particular feels quite important, as otherwise there's seemingly no way of restricting what addresses local users can use.

PR #779, opened 2017-04-30
Support 'senders' and 'received-auth' in 'listen on socket'

Wolf480pl:
A proof-of-concept implementation of #709

- make the lookup process offer a uid to userinfo lookup by means of IMSG_SMTP_LOOKUP_UID_USERINFO
  (I wasn't sure if this should go through a table or not, so chose the simpler option for now)
- add an optional username argument to smtp_session
- use the uid received by smtp process in IMSG_CTL_SMTP_SESSION
  to find the user's username, and pass it to the smtp_session
- respect the F_RECEIVEDAUTH flag whenever a user is authenticated,
  even if the connection is not SF_SECURE
- allow using `received-auth` and `senders` in `listen on socket` in the config


poolpOrg:
The feature request still exists but the PR was not acceptable back then and no longer applies, I'll close it and see if I can implement that differently.

For a back story, a lengthy discussion between @ericfaurot and I happened with regard to socket listeners, how they were implemented and how they should be implemented, a bit of the discussion provides insight on why your PR required special handling with regard to a network listener.

I'll tackle your feature request when we refactor a bit "listen on socket"

PR #780, opened 2017-04-30
Fix table_dump_lookup's default case

Wolf480pl:
Right now, if table_dump_lookup is called with a lookup type it cannot handle, eg. K_MADDRMAP,
it will return its static buffer with the contents from previous lookup, which can be very confusing, eg. when I see in the logs a senders table returning a password hash.

This PR fixes it by writing `"???"` to the buffer in the type switch reaches `default`

### DIFF
--- a/smtpd/table.c
+++ b/smtpd/table.c
@@ -702,6 +702,7 @@ table_dump_lookup(enum table_service s, union lookup *lk)
 		break;
 
 	default:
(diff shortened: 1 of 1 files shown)


poolpOrg:
I agree with your PR, however this doesn't fix the original problem which is that table_dump_lookup() should really have a case for K_MADDRMAP.

I'll merge your PR but can you also add the missing case ?

PR #781, opened 2017-05-01
Add K_MAILADDRMAP support to table_dump_lookup

Wolf480pl:
If there's too many addresses to fit in a line (because eg. someone didn't use whitespace in the table), the code puts "..."  at the end of the truncated string.

(in response to #780)


poolpOrg:
Sorry for the delay

Can you fix this diff to use strlcat() rather than the snprintf() loop with the written offset ?

Wolf480pl:
Updated the PR. Not sure if that's what you meant by using strlcat().

PR #802, opened 2017-07-28
Fix unstrutable error messages with missing mail spool directory

CasperVector:


poolpOrg:
Hi, care to sent these as two different PR ?

The getlock() one was basically okayed by @ericfaurot  and I, whereas the mda_getlastline() needs to be fixed and @ericfaurot would like to comment on it :-)

CasperVector:
Sorry, I forgot to separate the two PRs.
Now done in #803 and #804, the former revised according to the comment on `sz`.

PR #803, opened 2017-07-29
mda_getlastline(): handle empty output correctly.

CasperVector:


poolpOrg:
@ericfaurot diff reads fine to me, does it look better to you ?

ericfaurot:
yes, ok.

poolpOrg:
I'll commit to the OpenBSD tree shortly, github will be synced

PR #820, opened 2018-01-06
Cryptutil

bsdsx:
import bcrypt_checkpass


poolpOrg:
This doesn't seem correct to me, crypt_checkpass() becomes incompatible with system's crypt() if host doesn't support bcrypt.

I think crypt_checkpass() should be a wrapper around the crypt() function, not bring its own crypt algorithm.

PR #824, opened 2018-01-17
Fix --with-maildir configure option

Duncaen:
Without this patch its not possible to cross compile OpenSMTPD.
`$maildir` is never set if `--with-maildir` is specified and errors with `cross compiling: use --with-maildir=/path/to/mail`.


poolpOrg:
that's definitely not correct.

care to elaborate what you're trying to achieve and why you came up with this solution ?

Duncaen:
Looking over it again I can see that my patch is wrong.

Using `--with-path-mbox` works, but if its not given the configure script tries to guess the path by building an executable to get `_PATH_MAILDIR` which does not work while cross compiling and results in the warning that `--with-maildir=` can be used (an unrecognized option).

There is some mixup with `with-maildir` and `with-path-mbox` but its just confusing and doesn't really break the build.

Issue #738, opened 2016-11-06
Fails to build from source with OpenSSL >= 1.1.0


ryanakca:
I've started working through the compile errors and have gotten through those involving `SSL_CTX`. Still remaining are those involving `X509_STORE_CTX` and `RSA_METHOD`.

Here are examples of failed build logs involving these errors:
[opensmtpd_6.0.2p1-1_amd64-2016-11-06T15:44:01Z.txt](https://github.com/OpenSMTPD/OpenSMTPD/files/574168/opensmtpd_6.0.2p1-1_amd64-2016-11-06T15.44.01Z.txt)
[opensmtpd_6.0.2p1-1_amd64-2016-11-06T16:14:22Z.txt](https://github.com/OpenSMTPD/OpenSMTPD/files/574169/opensmtpd_6.0.2p1-1_amd64-2016-11-06T16.14.22Z.txt)

ryanakca:
Here's what I have so far. I make no guarantees on correctness, but it's what I've managed to put together by looking at the manpages and the OpenSSL source. It's also lacking any checks on the OpenSSL version, and so likely will FTBFS on systems with OpenSSL < 1.1.0.

See https://github.com/OpenSMTPD/OpenSMTPD/commit/8e42cd2b72fbc0cdc978d1c593f221da2af31306 .

poolpOrg:
This is not as simple as this :-/

OpenSSL made an API change without any transition plan which means that if we don't ship this, we can't build on OpenSSL 1.1.0 and if we ship this, we can't build on LibreSSL and any system that isn't shipping OpenSSL 1.1.0.

At the very least there should be compat code to deal with the accessors being missing but this is yet again an OpenSSL fucking everyone up without prior notice issue...

We need to figure out the proper way to deal with this and I'll have to check with the other OpenBSD hackers how they plan to cope with this in OpenSSH and other portable daemons.

landryb:
OpenSSL 1.0.x won't go away, and should still be installable in parallel of 1.1.x on decent systems. An option is to explicitely unsupport 1.1.x until the openssl guys fix their mess.

Vanhecke:
What about using LibreSSL-portable?

dfateyev:
The same issue with the latest build on Fedora devel (where openssl-1.1.0e is used).
Please refer [18270311 build log](https://kojipkgs.fedoraproject.org//work/tasks/311/18270311/build.log) for more details.

[...]

PR #825, opened 2018-01-19
Add support for libssl1.1 while preserving libressl and libssl 1.0.x support

ryanakca:
Patch by Sebastian Andrzej Siewior <sebastian@breakpoint.cc>, with description:

    SSL_F_SSL_CTX_USE_CERTIFICATE_FILE and
    SSL_CTX_clear_extra_chain_certs was around in openssl since before
    they forked it. So with this patch it should work with their
    libressl, libssl 1.0.2 and 1.1.

I updated this to apply to 6.0.3p1. I can't vouch for correctness---this is outside my realm of expertise---but it compiles and runs with both libssl 1.0.2n and 1.1.0g. Fixes issue #738.

Source: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=859544

### DIFF
--- a/openbsd-compat/libressl.c
+++ b/openbsd-compat/libressl.c
@@ -81,14 +81,14 @@ SSL_CTX_use_certificate_chain(SSL_CTX *ctx, char *buf, off_t len)
 	x = ca = NULL;
 
 	if ((in = BIO_new_mem_buf(buf, len)) == NULL) {
-		SSLerr(SSL_F_SSL_CTX_USE_CERTIFICATE_CHAIN_FILE, ERR_R_BUF_LIB);
+		SSLerr(SSL_F_SSL_CTX_USE_CERTIFICATE_FILE, ERR_R_BUF_LIB);
 		goto end;
 	}
 
 	if ((x = PEM_read_bio_X509(in, NULL,
-		    ctx->default_passwd_callback,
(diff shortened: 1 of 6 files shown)


poolpOrg:
Hi,

I can't vouch for correctness either, it reads fine but I need to get this reviewed by other openbsd hackers too, so I'd say if people at debian trust it you should package with it until we sort this out on our side.

The PR will likely be rearranged differently and not merged as is because it introduces ifdefs within .c files which we try to avoid unless there's absolutely no other way and given the diff within ca.c I think we can probably solve in a way that's ok for master & portable without introducing ifdef's in the src/ directory.

poolpOrg:
As mentionned in ticket #738, we have validated that we can build OpenSMTPD with LibreSSL on various distros including Ubuntu, Debian, CentOS, Fedora and Archlinux, so we now consider LibreSSL to be our target SSL library.

The reasons preventing LibreSSL and OpenSSL from co-existing are no longer relevant, there is no reason for our small team to keep maintaining two different SSL libraries given that our ultimate goal is to switch to LibreSSL's smaller libtls to simplify our code base and avoid making mistakes with the complexity of the libssl API.

Distributions willing to retain OpenSSL dependency are free to do so, they'll just get to deal with the extra work  :-)

PR #829, opened 2018-02-05
defines.h: Add openbsd-compat define for CLOCK_REALTIME

twinshadow:
MacOS 10.10 does not implement clock_gettime(), and so is missing the defines
for CLOCK_REALTIME and CLOCK_MONOTONIC. A definition exists for CLOCK_MONOTONIC
in openbsd-compat , but CLOCK_REALTIME was missing.

    ../../smtpd/queue_fs.c:763:20: error: use of undeclared identifier 'CLOCK_REALTIME'
            if (clock_gettime(CLOCK_REALTIME, &startup))
                              ^
    1 error generated.
    make[3]: *** [../../smtpd/smtpd-queue_fs.o] Error 1

### DIFF
--- a/openbsd-compat/defines.h
+++ b/openbsd-compat/defines.h
@@ -475,6 +475,10 @@ typedef uint16_t	in_port_t;
 #ifndef CLOCK_MONOTONIC
 #define CLOCK_MONOTONIC -1
 #endif
+
+#ifndef CLOCK_REALTIME
+#define CLOCK_REALTIME -2
+#endif
 /* end of chl */
 
 #ifndef HAVE_FPARSELN


twinshadow:
I am not happy with this solution, since the defines for CLOCK_MONOTONIC and CLOCK_REALTIME are ignored by the openbsd-compat `clock_gettime()`. Right now, the differences do not matter between the two. CLOCK_REALTIME is only called at `fs_queue_init()` and all comparisons are with filesystem modified-time. If something were to change in the future where CLOCK_REALTIME was expected to act and be compared to other CLOCK_REALTIME values, then this PR would be a liability.

poolpOrg:
maybe we should question the need for a clock_gettime(CLOCK_REALTIME) for this particular use case...

I'll merge this because it improves wrt today but I think we'll have to reassess the use of counters within opensmtpd, it's a bit brittle as is for no particular gain.

Issue #814, opened 2017-10-11
Message-Id header not added if missing and listener port is 587


greecemunky:
In a new setup (OS up-to-date Debian testing, OpenSMTPD 6.0.2p1 from Debian packages), Message-Id is not being set.

I started fresh, compiling OpenSMTPD 6.0.2p1 from source to debug.  The following needs to be changed.

smtpd/smtp_session.c

From:
```if (s->listener->local || s->listener->port == 587) {```

To:
```if (s->listener->local || ntohs(s->listener->port) == 587) {```

Works as expected now.  I also tested the amended code against port 25 and the header is still not present (also as expected).

FWIW, I don't know C, I just saw ```ntohs(listener->port)``` elsewhere and thought I'd give it a try.  I also am not familiar with push/pulling etc. and don't want to do anything silly so thought I'd raise it here as an issue instead.

Also, can/should the "missing header" check for Date be moved outside the if block?

**Comment by greecemunky** [comment]
Patches against current master branch:

If keeping "missing header" check for Date within if block:

```
--- smtpd/smtp_session.c_orig   2017-09-19 18:20:13.000000000 +1000
+++ smtpd/smtp_session.c        2017-10-12 10:03:08.756026156 +1100
@@ -1234,7 +1234,7 @@
        rfc2822_body_callback(&tx->rfc2822_parser,
            dataline_callback, s);

-       if (s->listener->local || s->listener->port == 587) {
+       if (s->listener->local || ntohs(s->listener->port) == 587) {
                rfc2822_missing_header_callback(&tx->rfc2822_parser, "date",
                    header_missing_callback, s);
                rfc2822_missing_header_callback(&tx->rfc2822_parser, "message-id",
```

Or, if moving "missing header" check for Date outside if block:

```
--- smtpd/smtp_session.c_orig   2017-09-19 18:20:13.000000000 +1000
+++ smtpd/smtp_session.c        2017-10-12 10:07:41.723574708 +1100
@@ -1233,10 +1233,10 @@
            header_domain_append_callback, s);
        rfc2822_body_callback(&tx->rfc2822_parser,
            dataline_callback, s);
+       rfc2822_missing_header_callback(&tx->rfc2822_parser, "date",
+           header_missing_callback, s);

-       if (s->listener->local || s->listener->port == 587) {
-               rfc2822_missing_header_callback(&tx->rfc2822_parser, "date",
-                   header_missing_callback, s);
[...]

PR #830, opened 2018-02-06
Fix Message-Id header not added if missing and listener port is 587

greecemunky:
#814 suggests two possible fixes, this PR covers the first suggestion (smallest change to existing code).

### DIFF
--- a/smtpd/smtp_session.c
+++ b/smtpd/smtp_session.c
@@ -1156,7 +1156,7 @@ smtp_tx(struct smtp_session *s)
 	rfc2822_body_callback(&tx->rfc2822_parser,
 	    dataline_callback, s);
 
-	if (s->listener->local || s->listener->port == 587) {
+	if (s->listener->local || ntohs(s->listener->port) == 587) {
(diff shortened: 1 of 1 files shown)


poolpOrg:
I don't think this is the correct way to fix this, or more precisely not the proper _place_ to fix this.

If we are going to manipulate the listener port in smtp_session, then it should not require its bits being reordered otherwise it's too easy to forget a place (how ironic that this is what you actually fix ... :p)

Currently the listener creation in parse.y has htons() everywhere a port is being passed but I think this is wrong, the port should always be in host order and only converted to network order whenever a call requires them (ie: when binding the socket) not when code performs logic that's not socket related.

Care to check how this could be fixed in the proper place ?

greecemunky:
I would like to, but my C knowledge is _very_ limited and any changes I'd think of will most likely break more than it'll fix.  Seriously, you don't want me working on anything major.  :P

Is it possible to accept this PR to resolve this issue and create another issue detailing rework of how ports are handled?

poolpOrg:
we have committed to OpenBSD a variant of your diff as a quickfix because OpenBSD release is so close, we'll commit real fix after release.

thanks

Issue #819, opened 2018-01-03
Missing support for RFC4954 AUTH LOGIN initial-response


biotecs:
Due to RFC4954 the "AUTH LOGIN" smtp command can be used with username as initial response. In this case the state should change to:
https://github.com/OpenSMTPD/OpenSMTPD/blob/4d34dfe9cffe5683574342d7bbef6a7c9de00e11/smtpd/smtp_session.c#L1688

and return without doing:

https://github.com/OpenSMTPD/OpenSMTPD/blob/4d34dfe9cffe5683574342d7bbef6a7c9de00e11/smtpd/smtp_session.c#L1689

In my opinion there should be something like:

```
smtp_enter_state(s, STATE_AUTH_USERNAME); 
if (strchr(arg, '\0') != NULL) {
    smtp_rfc4954_auth_login(s, arg);
    return;
}
smtp_reply(s, "334 VXNlcm5hbWU6"); 
```
In 2014 there was an accepted pull request already: https://github.com/OpenSMTPD/OpenSMTPD/pull/425. But it is not existent in master anymore.

If you guys can confirm this, I would create a new pull request.

b01t:
This breaks the way the .Net SmtpClient authenticates, meaning a bunch of tools written in .Net can not authenticate with this server. Please (re-) fix? :)

poolpOrg:
fix committed to OpenBSD, will be synced to github shortly.

PR #834, opened 2018-02-20
Fixing support for RFC4954 AUTH LOGIN initial-response

biotecs:
Due to RFC4954 the "AUTH LOGIN" smtp command can be used with username as initial response. In this case the state should change to:
https://github.com/OpenSMTPD/OpenSMTPD/blob/4d34dfe9cffe5683574342d7bbef6a7c9de00e11/smtpd/smtp_session.c#L1688

and return without doing:

https://github.com/OpenSMTPD/OpenSMTPD/blob/4d34dfe9cffe5683574342d7bbef6a7c9de00e11/smtpd/smtp_session.c#L1689

In my opinion there should be something like:

```
smtp_enter_state(s, STATE_AUTH_USERNAME); 
if (strchr(arg, '\0') != NULL) {
    smtp_rfc4954_auth_login(s, arg);
    return;
}
smtp_reply(s, "334 VXNlcm5hbWU6"); 
```
In 2014 there was an accepted pull request already: https://github.com/OpenSMTPD/OpenSMTPD/pull/425. But it is not existent in master anymore.

I already opened an issue here: https://github.com/OpenSMTPD/OpenSMTPD/issues/819

### DIFF
--- a/smtpd/smtp_session.c
+++ b/smtpd/smtp_session.c
@@ -1686,6 +1686,10 @@ smtp_rfc4954_auth_login(struct smtp_session *s, char *arg)
 	switch (s->state) {
 	case STATE_HELO:
 		smtp_enter_state(s, STATE_AUTH_USERNAME);
+		if (strchr(arg, '') != NULL) {
+			smtp_rfc4954_auth_login(s, arg);
+			return;
+		}
 		smtp_reply(s, "334 VXNlcm5hbWU6");
 		return;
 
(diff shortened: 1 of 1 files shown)


poolpOrg:
there is something wrong with your diff but a variation will be committed shortly

poolpOrg:
diff committed to OpenBSD, github will be synch-ed shortly.

closing this merge request and ticket #819

PR #845, opened 2018-04-09
Allow addresses with prefix lengths in SPF records

ryanakca:
Revision 1.8 to src/usr.sbin/smtpd/spfwalk.c removed support for prefix lengths in records. In particular, it was returning nothing as output for the SPF record for "gmail.com" because each entry in gmail's SPF record has a prefix length. This commit reintroduces support for prefix lengths.

I factored the check out into a separate function, because it was becoming unreadable when inlined into each conditional. This diff should apply (modulo some fuzz) against OpenBSD's CVS: http://www.cs.cmu.edu/~rkavanag/spfwalk.diff

### DIFF
--- a/smtpd/spfwalk.c
+++ b/smtpd/spfwalk.c
@@ -42,6 +42,7 @@ static void	dispatch_mx(struct dns_rr *);
 static void	dispatch_a(struct dns_rr *);
 static void	dispatch_aaaa(struct dns_rr *);
 static void	lookup_record(int, const char *, void (*)(struct dns_rr *));
+static int	validate_address(int, const char *);
 static void	dispatch_record(struct asr_result *, void *);
 static ssize_t	parse_txt(const char *, size_t, char *, size_t);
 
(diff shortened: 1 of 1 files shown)


smtpd/spfwalk.c
```diff
@@ -129,6 +130,17 @@ dispatch_record(struct asr_result *ar, void *arg)
 	}
 }
 
+int
+validate_address(int af, const char *s)
+{
+	struct in6_addr ina;
+
+	if (strrchr(s, '/') != NULL)
+		return inet_net_pton(af, s, &ina, sizeof(s)) != -1;
```
snimmagadda: The size argument to inet_net_pton(3) should be the size of dst buffer.

smtpd/spfwalk.c
```diff
@@ -129,6 +130,17 @@ dispatch_record(struct asr_result *ar, void *arg)
 	}
 }
 
+int
+validate_address(int af, const char *s)
+{
+	struct in6_addr ina;
+
+	if (strrchr(s, '/') != NULL)
```
snimmagadda: Do we need to check explicitly check for prefix notation? How about... https://nimmagadda.net/walk.diff

ryanakca: Good catch on the wrong size argument to inet_net_pton, thanks. Your patch looks much cleaner than mine, and rereading the inet_net_pton(3) man page, I think it should work (though I haven't tested it). You may want to consider applying an analogous change to the function `host_v4` in `src/pfctl/pfctl_parser.c` (lines 1747--1753) off of which my patch was based.

poolpOrg:
merged

PR #847, opened 2018-04-12
add brackets around Return-Path as per RFC5322 §3.6.6

Ekleog:
The absence of <> breaks LDA with the dovecot 2.3.0.1 -> 2.3.1 upgrade, even though dovecot developers have told me it'll be made more lenient in the next update.

This should fix the RFC-non-compliance from OpenSMTPD side. :)

Thanks for your work, as always!


poolpOrg:
merged

PR #849, opened 2018-04-12
Make Return-Path RFC5322 compliant

### LINKED ISSUE #848: Return-Path not rfc5322 compliant
Currently, OpenSMTPd just adds
```
Return-Path: user@example.org
```
to messages for mda. This trips up Dovecot 2.3.1, see https://www.dovecot.org/pipermail/dovecot/2018-April/111570.html

Standard indicates that the mail-address should appear in angled brackets.
From https://tools.ietf.org/html/rfc5322#section-3.6.7
```
return          =   "Return-Path:" path CRLF
path            =   angle-addr / ([CFWS] "<" [CFWS] ">" [CFWS])
```

Dovecot is fixing this, but probably better to add to OpenSMTPd too.

Sp1l:
From https://tools.ietf.org/html/rfc5322#section-3.6.7
```
return          =   "Return-Path:" path CRLF
path            =   angle-addr / ([CFWS] "<" [CFWS] ">" [CFWS])
```

Fixes  #848

### DIFF
--- a/smtpd/mda.c
+++ b/smtpd/mda.c
@@ -266,7 +266,7 @@ mda_imsg(struct mproc *p, struct imsg *imsg)
 				 * if any
 				 */
 				n = io_printf(s->io,
-				    "Return-Path: %s\n"
+				    "Return-Path: <%s>\n"
 				    "Delivered-To: %s\n",
 				    e->sender,
 				    e->rcpt ? e->rcpt : e->dest);


Ekleog:
For the record, had opened https://github.com/OpenSMTPD/OpenSMTPD/pull/847 just after reading Stephan's answer, looks like we both had the same reflex :p

Sp1l:
Duplicate of #847

PR #858, opened 2018-05-31
Fix build portable

bsdsx:
- remove LINE_MAX (not used)
- pledge is defined in openbsd-compat.h


poolpOrg:
i dont think this is right, no

i think the problem is that we lack the 

    #include "includes.h"

which is the first header to include in every .c file within smtpd/, look at other files

PR #861, opened 2018-06-15
Don't call err with null param

bsdsx:
Macro err don't accept concat string with null


poolpOrg:
that doesn' t seem right, the err() macro should be fixed if it doesn't support err(1, NULL) because it is very common...

poolpOrg:
in addition, this means i have to apply this change within OpenBSD to avoid a delta with portable and it makes no sense there :-/

bsdsx:
wrong pr
fixed with bsd-err.c

PR #865, opened 2018-06-18
parse.y don't use any _PATH_*

bsdsx:


bsdsx:
#include <util.h> seems useless

poolpOrg:
sorry, forgot about this one, I'll have to check if this can _also_ be safely removed from openbsd or if it'll create a delta with portable.

will get back to you this week-end !

PR #876, opened 2018-09-05
minor linux portability fixes

rkitover:
Linux/glibc does not have O_EXCL for open, sockaddr.sa_len,
sockaddr_storage.ss_len or the EFTYPE errno.

Use autoconf macros to conditionally compile replacements.

If EFTYPE is unavailable, replace it with EMEDIUMTYPE which seems to be
the closest.

Also check for automake 16.

These changes were necessary for OpenSuse Tumbleweed, these changes
should also apply to many other distributions.

We are running this code for a server and it seems to be working
correctly.

If, for the sake of de-uglification, you would like to just use the more
portable code for any of these changes without the conditional
compilation, I would be happy to make this change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -137,7 +137,6 @@ AC_CHECK_HEADERS([ \
 	sys/cdefs.h \
 	sys/dir.h \
 	sys/file.h \
-	sys/mount.h \
 	sys/ndir.h \
 	sys/pstat.h \
 	sys/statfs.h \
@@ -357,6 +356,16 @@ AC_CHECK_DECLS([O_NONBLOCK], [], [], [
 #endif
 ])
 
+AC_CHECK_DECLS([O_EXCL], [], [], [
+#include <sys/types.h>
+#ifdef HAVE_SYS_STAT_H
+# include <sys/stat.h>
+#endif
+#ifdef HAVE_FCNTL_H
+# include <fcntl.h>
+#endif
+])
+
 AC_CHECK_DECLS([AF_LOCAL, PF_LOCAL], [], [], [
 #include <sys/socket.h>
 ])
@@ -375,6 +384,11 @@ AC_CHECK_DECLS([LOG_PERROR], [], [], [
 #include <syslog.h>
 ])
 
+AC_CHECK_DECLS([EFTYPE], [], [
+	AC_MSG_ERROR([please define EFTYPE, by e.g. passing -DEFTYPE=EMEDIUMTYPE in your CFLAGS])
+], [
+#include <errno.h>
+])
 
 #
 # CHECKS FOR COMPILER CHARACTERISTICS
--- a/smtpd/mail.mboxfile.c
+++ b/smtpd/mail.mboxfile.c
@@ -14,9 +14,14 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#include <sys/types.h>
-#include <sys/stat.h>
+#include "config.h"
 
+#include <sys/types.h>
+
+#include <sys/stat.h>
+#include <sys/file.h>
+
+#include <time.h>
 #include <ctype.h>
 #include <err.h>
 #include <errno.h>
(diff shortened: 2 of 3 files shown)


bsdsx:
for socket len: https://github.com/OpenSMTPD/OpenSMTPD/pull/878
for EFTYPE, ./configure -DEFTYPE= ?

poolpOrg:
so i have merged the fix from @bsdsx with regard to sockaddr and committed to your pull request.

im ok with the configure layer stuff but not happy with the changes within smtpd/, will comment inline

configure.ac
```diff
@@ -140,6 +140,7 @@ AC_CHECK_HEADERS([ \
 	sys/mount.h \
 	sys/ndir.h \
 	sys/pstat.h \
+	sys/stat.h \
```
poolpOrg: do we really need to check for sys/stat.h ?

it is posix, exists even on windows and i've never seen a Linux that doesn't have it

configure.ac
```diff
@@ -357,6 +358,16 @@ AC_CHECK_DECLS([O_NONBLOCK], [], [], [
...
+AC_CHECK_DECLS([O_EXCL], [], [], [
+#include <sys/types.h>
+#ifdef HAVE_SYS_STAT_H
+# include <sys/stat.h>
+#endif
+#ifdef HAVE_FCNTL_H
+# include <fcntl.h>
+#endif
+])
+
```
poolpOrg: HAVE_SYS_STAT_H set aside, ok with this

openbsd-compat/defines.h
```diff
@@ -418,6 +418,10 @@ typedef uint16_t	in_port_t;
 # define EWOULDBLOCK EAGAIN
 #endif
 
+#ifndef EFTYPE
+# define EFTYPE EMEDIUMTYPE
+#endif
+
```
poolpOrg: ok by me but I think @ericfaurot should have used another errno, we can re-assess that later, EFTYPE is only used in one particular case and your fix is acceptable.

smtpd/mail.mboxfile.c
```diff
@@ -14,13 +14,24 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include "config.h"
+
 #include <sys/types.h>
+
+#ifdef HAVE_SYS_STAT_H
```
poolpOrg: unsure about this.

either way, if there's a real reason to test sys/stat and do conditional inclusion, please do not reorder headers as this makes diffing against the non-portable version trickier.

The same comment applies to all headers that are reshuffled in this file.

smtpd/mail.mboxfile.c
```diff
@@ -71,10 +82,19 @@ mboxfile_engine(const char *sender, const char *filename)
 
 	time(&now);
 
-	fd = open(filename, O_CREAT | O_APPEND | O_WRONLY | O_EXLOCK, 0600);
+#ifdef HAVE_DECL_O_EXCL
```
poolpOrg: we avoid ifdef as much as possible in smtpd, i suggest we discuss this separately from the other diffs which are mostly ok minus minor fixes.

openbsd-compat/defines.h
bsdsx: On OS without EFTYPE, packager should add '-DEFTYPE=ESOMETHING' on CFLAGS

poolpOrg: fair enough, either way I think the _real_ fix would be to pick another errno since this one is not really the best choice for that particular use-case it was used for.

@rkitover I agree with bsdsx, using a define to redefine it would be less intrusive of a hack

smtpd/mail.mboxfile.c
rkitover: I will remove these.

configure.ac
rkitover: I think autoconf checks for it automatically so I can remove this.

openbsd-compat/defines.h
rkitover: The packager certainly can do this, but this will put a lot of people off who are just trying to compile the app on linux and use it and may not understand how to do this.

I think some kind of automatic configure thing would be good, but up to you.

smtpd/mail.mboxfile.c
rkitover: I can find/make a macro solution for this if you like.

rkitover: Or maybe a function in the openbsd compat part, such as `open_locked_atomic()` or something.

poolpOrg:
I have committed the separate bits, thanks !

PR #882, opened 2018-10-13
fix strndup

bsdsx:
next commit fix strnlen


openbsd-compat/strnlen.c
```diff
@@ -0,0 +1,33 @@
...
+strnlen(const char *str, size_t maxlen)
+{
+	const char *cp;
+
+	for (cp = str; maxlen != 0 && *cp != '\0'; cp++, maxlen--)
+		;
+
+	return (size_t)(cp - str);
+}
+DEF_WEAK(strnlen);
```
poolpOrg: DEF_WEAK should probably be stripped

PR #883, opened 2018-10-23
bootstrap: allow automake 1.16

Ekleog:
This change allows running `bootstrap` with `automake-1.16`, which appears to allow successful building (provided the right changes are made, cf. message just below)

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -36,7 +36,7 @@ aclocalflags="`sed -ne 's/^[ \t]*ACLOCAL_AMFLAGS[ \t]*=//p' Makefile.am 2>/dev/n
 
 # Check for automake
 amvers="no"
-for v in 15 14; do
+for v in 16 15 14; do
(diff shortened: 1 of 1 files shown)


Ekleog:
Actually nvm, I completely failed… well, this patch is still required, though

Ekleog:
So to copy information from IRC over here, to make builds and tests pass successfully there's need for an additional patch (available on IRC, as I don't own copyright on all of it, what I've written is written with the aim of being included in OpenSMTPD under the proper license).

With this additional patch, our build and test both pass. However, I must say lines like [this one](https://github.com/OpenSMTPD/OpenSMTPD/blob/portable/smtpd/parse.y#L594) make packaging it require a `sed` over those lines, because NixOS doesn't put `libexec` at this position. I don't know if you'd be willing to add a compile-time configuration parameter to set this, though.

poolpOrg:
sorry this got fixed in parallel as I was fixing build for ubuntu.

thanks for the PR though

PR #884, opened 2018-10-27
Add missing object file for smtpctl

Ekleog:
With this change and addition of `CFLAGS=-ffunction-sections` and
`LDFLAGS=-Wl,--gc-sections`, it now builds.

`purge_config` is defined in `config.c` but used in `parse.y`, making
`config.c` mandatory. But `config.c` requires lots of functions from
other files in other unrelated functions, meaning that these unrelated
(and unused in `smtpctl`) functions need to be garbage-collected.

This is a workaround, a better way would likely be to refactor
`purge_config` and `config_default` out of `config.c`, but I don't want
to push for a change that would separate the portable version and the
OpenBSD version too much.

### DIFF
--- a/mk/smtpctl/Makefile.am
+++ b/mk/smtpctl/Makefile.am
@@ -4,6 +4,7 @@ sbin_PROGRAMS=		smtpctl
 
 smtpctl_SOURCES=	$(smtpd_srcdir)/enqueue.c
 smtpctl_SOURCES+=	$(smtpd_srcdir)/parser.c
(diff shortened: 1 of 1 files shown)


poolpOrg:
definitely not correct.

parse.c is generated from parse.y, can you explain what prompted you to add parse.c to the list ?

Ekleog:
Oh. Definitely not correct indeed, I failed copying the patch into a PR…
This was supposed to be adding `config.c`, not `parse.c`.

I've pushed an updated PR, and now the PR contents is compatible with
the commit message / PR text :)

PR #895, opened 2018-12-21
fix cross compilation

AlexandreFournier:
AC_RUN_IFELSE is missing the last argument [action-if-cross-compiling]

See also: https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Runtime.html


poolpOrg:
Thanks !

PR #896, opened 2018-12-27
OpenSMTPD - add warning when trying to deliver to root user with maildir method

gunhu:


poolpOrg:
nope, this is not correct for multiple reasons, but the two main are:

- queue.c is not where delivery takes place so this is not where this check should happen
- delivery to any privileged user is disallowed so you can't just test "root"

The relevant bit of code is:

https://github.com/OpenSMTPD/OpenSMTPD/blob/master/smtpd/smtpd.c#L1379-L1380

Where it should be a log_warn()

poolpOrg:
I have taken care of this using a different approach, closing the PR.

Issue #812, opened 2017-08-30
extras: table-ldap - bind with user credentials given during smtp auth


pvagner:
I am trying to setup opensmtpd + dovecot + openldap email system on arch linux.
This blog article from 2015 is the only usefull source of info I was able to gather on this topic. I mean this one: https://www.tumfatig.net/20150718/opensmtpd-dovecot-and-ldapd-on-openbsd-5-7/
After experimenting a bit with similar config featured in the article and finally finding this unanswered post from september 2016: https://www.mail-archive.com/misc@opensmtpd.org/msg03168.html
I have came to some assumptions:
* OpenSMTPd binds to ldap by using the credentials from the config file e.g. /etc/mail/ldap.conf . It newer hands smtp credentials to ldap server. It retrieves credentials according to the credentials_filter and credentials_attributes configuration options.
* So by knowing OpenSMTPd does not bind to ldap with user provided credentials I think we have to configure the same password hashing scheme for opensmtpd and openldap. This is kinda unfortunate as this is something ldap is supposed to address and we should not have to wory about password schemes, hashing and similar stuff.
* So at one hand we do have OpenSMTPd with its built-in non-configurable crypt 3 password hashing (on my system these are SHA512 hashes, the crypt implementation and configuration may differ I think) for example...
    # smtpctl encrypt mysupersecretpassword
    $6$tEwwZAWy.bLlIDK0$27WDpDFlCGjIP8pl6aTIazkirsG.f9Ux.Djnl4uZYcgWUsl68ID9q.3k5ndr5AMVYSc1tmSPawaX//OxJHk2e1
* At the other hand we do have openldap which defaults to {SSHA} hashing for example
    # slappasswd -s mysupersecretpassword
    {SSHA}ysPSlo0fJBmbVGQ7B9eJ4+eiuCSLT2XD
* There is some configurability at the openldap side of things that we can configure openldap to use {CRYPT} and configure SHA512 .
I don't have a guide here but this is pretty close I believe... http://www.openldap.org/lists/openldap-technical/201408/msg00098.html
* Conclusion even if we configure openldap to store its passwords using SHA512 crypt 3 based hashes similar to how OpenSMTPd generates passwords we will still need some changes as the resulting strings are not easily comparable. We would need to strip {CRYPT} prefix from the openldap password and we would have to remove ///randomsalt from the opensmtpd password.
[...]

PR #903, opened 2019-01-28
Allow to check user credentials in the table backend

cdidier:
Some backend doesn't use crypt to store the passwords so we
cannot just fetch the password with K_CREDENTIALS because
it might be hashed with an unknown method. We need another
way to authenticate the user directly in the table backend.

- Introduce a new table service K_CHECKUSER which can be
  used to check the credentials in the table backend.
  The key must have the form "user:password".
- Make sure the key is never converted to lowercase when using
  K_CHECKUSER (we don't want the password to be converted to
  lowercase).
- When authenticating a new smtp session, first try to use
  K_CHECKUSER to authenticate using the backend, and then if
  it fails use K_CREDENTIALS and the crypt_checkpass(3) to keep
  backward compatibility.

### DIFF
--- a/smtpd/lka.c
+++ b/smtpd/lka.c
@@ -689,6 +689,8 @@ lka_authenticate(const char *tablename, const char *user, const char *password)
 {
 	struct table		*table;
 	union lookup		 lk;
+	char			*buf;
+	int			 buflen, r;
 
 	log_debug("debug: lka: authenticating for %s:%s", tablename, user);
 	table = table_find(env, tablename);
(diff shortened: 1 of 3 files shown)


cdidier:
With https://github.com/OpenSMTPD/OpenSMTPD-extras/pull/55, its a first try to solve https://github.com/OpenSMTPD/OpenSMTPD/issues/812.

poolpOrg:
Not rejecting this pull request yet but we should discuss this further because I'm not fond of the idea of a new lookup service to solve this issue.

I haven't dived into your issue yet so I don't know if the solution is at the lookup services level, but if it were I'd rather investigate how it can be solved by updating K_CREDENTIALS which was created for the purpose of looking up credentials, rather than add a new lookup service that would do the same but with one difference.

I'll move the discussion to the issue you created.

poolpOrg:
My comments in issue #812 were not addressed/discussed and the PR no longer applies so I'll close it for now, we can reopen if this becomes active again.

Issue #905, opened 2019-02-19
Unable to build Portable, table.c:66:3 error: 'table_backend_db' undeclared here


EmperorArthur:
Hello,

I am having trouble building OpenSMTPD. I think there's an issue where an optional feature was accidentally made a requirement.  It probably wasn't caught because most developers working on the project have the option installed.

# The issue
I am unable to build the latest portable branch because the code references "table_backend_db", which is only added if **HAVE_DB_API** is defined.  This is vexing because `./configure` does not error if that is not present, and I'm unsure of what package I need to build/install to fix it.

I suspect this is a bug because I can build the "opensmtpd-6.4.1p2" tag successfully, and the requirements in the Readme have not changed.

# Solution 1:
Modify the code so that it no longer relies on **HAVE_DB_API**.  This is a silent requirements change that may have broken any build system that followed the directions in the Readme.

# Solution 2:
Modify 'configure.ac' to require **HAVE_DB_API** and provide a useful warning about what is missing if that is not found.  Further, update the Readme with the new requirements.

poolpOrg:
hi, thanks for your report

care to pull and give another try ?

EmperorArthur:
Confirmed fixed.  Thanks for the quick reply.

PR #906, opened 2019-02-19
Add a Dockerfile

EmperorArthur:
This allows for running OpenSMTPD in a Docker container.

Note: Because of issue #905 this will not build.

Once that is fixed, it should be pretty easy to have a CI pipeline build the container as a sanity check for any future pull requests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+FROM alpine:3.9 as build
+RUN apk add --no-cache \
+    ca-certificates \
+    wget \
+    cmake \
+    automake \
+    autoconf \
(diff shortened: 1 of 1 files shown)


EmperorArthur:
Codacy's still going to complain about removing '/usr/local/', but that's intended behavior in this case.  It's populated by a few empty folders, and `make install` re-creates it with only the OpenSMTPD code.

With #905 resolved, this should be a pretty easy CI test to make sure it never happens again.

poolpOrg:
I'm not well versed in Docker but I guess we can improve in tree so no objection from me.

PR #907, opened 2019-02-20
Switch Dockerfile to using /etc/mail/ for all configuration

EmperorArthur:
Some versions of docker have problems mounting single files.
Also include directions for using it in the Readme.

Some of the man pages refer to '/etc/mail/smtpd.conf', others to '/etc/smtpd.conf', which is the program's default, and `./configure` defaults to '/usr/local/etc/smtpd.conf'. I'm using '/etc/smtpd/smtpd.conf' because many programs seem to use the smtpd folder.

If you'd like this in a different folder, for example '/etc/mail' it's an easy change.  However leaving the file in just '/etc' wasn't working with docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ COPY . /opensmtpd
 #build opensmtpd
 RUN rm -r /usr/local/
 RUN ./bootstrap && \
-    ./configure --with-gnu-ld --sysconfdir=/etc --with-path-empty=/var/lib/opensmtpd/empty/ && \
+    ./configure --with-gnu-ld --sysconfdir=/etc/mail --with-path-empty=/var/lib/opensmtpd/empty/ && \
     make && \
     make install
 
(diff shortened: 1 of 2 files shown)


poolpOrg:
I think we should use /etc/mail which is the non-configurable default in the native version, the mess in sysconfdir comes from the variety of different configuration across different BSD systems and Linux distributions.

As a side-note, the man pages are patched during install so you shouldn't see a mix of /etc/mail and /etc as they are supposed to all point to the correct place. The only exceptions to this rule are paths to `mailname` and `resolv.conf` (not even sure if mentioned) which are historical and smtpd will not assume them to be in its own sysconfdir.

I'm ok with this merge request but /etc/mail would be less weird to me :-p

EmperorArthur:
Sure, that's an easy fix to make.

poolpOrg:
merged thanks !

Issue #908, opened 2019-02-20
Regression: TLS certificates cause immediate crashes in Portable branch (Docker)


EmperorArthur:
I was unable to run the latest portable branch in a Docker container when using tls.  Fortunately, the latest release (opensmtpd-6.4.1p2) does work.  That's why I've marked this as a regression.

Grepping through the code, the fatal error is called by smtpd.c:205 because `imsg === NULL`.  The debug statement then led me to mproc.c:172 when the pipe is closed.  I've confirmed it's the EV_READ case not the EV_WRITE.  That's as far as I was able to go unfortunately.

# Steps to reproduce
Use the provided smtpd.conf file with an empty '/etc/smtpd/passwd' file, and a valid TLS certificate/key.

## smtpd.conf
> pki_domain = "mail.cd-net.net"
> pki $pki_domain cert "/etc/ssl/private/sites/fullchain.cer"
> pki $pki_domain key  "/etc/ssl/private/sites/site.key"
> 
> table passwords file:/etc/smtpd/passwd
> 
> listen on localhost
> listen on eth0          tls         hostname $pki_domain pki $pki_domain auth-optional <passwords>
> listen on eth0 port 465 smtps       hostname $pki_domain pki $pki_domain auth-optional <passwords>
> listen on eth0 port 587 tls-require hostname $pki_domain pki $pki_domain auth          <passwords>
> 
> action "relay" relay helo $pki_domain
> match auth from any for any action "relay"

## Error Log (run with `-d -v -P mda -T all`)
> debug: init ssl-tree
> info: loading pki information for mail.cd-net.net
> debug: init ca-tree
> debug: init ssl-tree
> info: loading pki keys for mail.cd-net.net
> debug: using "fs" queue backend
> debug: using "ramqueue" scheduler backend
> debug: using "ram" stat backend
> info: OpenSMTPD 6.4.0-portable starting
> debug: init ssl-tree
> info: loading pki information for mail.cd-net.net
> debug: init ca-tree
> debug: init ssl-tree
> info: loading pki keys for mail.cd-net.net
> debug: using "fs" queue backend
> debug: using "ramqueue" scheduler backend
> debug: init ssl-tree
> info: loading pki information for mail.cd-net.net
> debug: init ca-tree
> debug: init ssl-tree
> info: loading pki information for mail.cd-net.net
> debug: init ca-tree
> debug: init ssl-tree
> info: loading pki keys for mail.cd-net.net
[...]

PR #911, opened 2019-02-21
Add an automated test to check if TLS certificates work

EmperorArthur:
This should allow Docker Hub to act as a CI platform.

NOTE:  This is testing for #908 which means Docker Hub's test should fail until it is resolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 *.in
 *.out
 *.log
-*~
 .#*
 .deps
 m4
@@ -26,6 +25,8 @@ stamp-h1
 ylwrap
 tags
 obj
+
+#Other VCS files
 CVS/
 smtpd/CVS
 smtpd/smtpctl/CVS
@@ -34,3 +35,7 @@ smtpd/mail/CVS
 smtpd/mail/CVS
 smtpd/mail/*/CVS
 smtpd/smtpctl/CVS/*
+
+#Editor temporary files
+*~
+.idea
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,9 @@ FROM alpine:3.9 as build
 
 WORKDIR /opensmtpd
 
+# libressl is used for testing only
 RUN apk add --no-cache \
     ca-certificates \
-    wget \
-    cmake \
     automake \
     autoconf \
     libtool \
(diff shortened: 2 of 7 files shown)


poolpOrg:
I love the idea however I'm not a big fan of having tests and regress in the main project repo because some of the regress tests are only going to apply to native (because there's an OpenBSD regress framework we're going to use, others only going to apply only to portable (ie: docker), others are going to apply to both (ie: the smtpscript utility from @ericfaurot which is already in its own repo).

I think we should create a specific repo which contains all the plumbing to run docker CI, this could very well be within the OpenSMTPD organisation.

What's your take on that ?

EmperorArthur:
Honestly not a fan.  I'll try to lay out my reasoning and where I see problems developing.  Looking at this it may come across as somewhat strong, but I just can't see a way to do it easily that ends well.  I think the real problem is that aside from two files these new tests don't use the regress framework.  Fixing that should solve things.

## CI / CD Issues
Pretty much every CI/CD platform I've worked with requires at least a configuration file in the main repository.  In the case of Docker Hub that's 'docker-compose.test.yml' and the Dockerfile.

It's, technically, doable, but you'd have to set up a custom webhook in Github pointing to your CI system and figure out how to deal with the almost 40 different hooks it sends to one URL.  Plus, it makes things like auto building from tags "fun".

## Docker doesn't like it (when building by hand)
Docker itself is designed with the file in the repository.  Docker actually won't go above the location where it's called from.  You can put the file itself in a subfolder and build with `docker build ./<folder name>/Dockerfile` but the only way to do that with another repository is to use git submodules, which makes things all sorts of complicated.

## Simultaneous releases suck
That's the Dockerfile and the CI file.  So, what about the tests themselves?  Those can be moved to a different repository, but you then run into the issues of having to do simultaneous deployments, and making sure users know about the second repository.

I faced a real life example of this from my previous employer.  They decided the backend and GUI were two different "microservices" with different repositories.  Yet, every time we changed one we had to make sure the other was changed at the same time.  The coordination even for a single developer just ate up time.  When it was teams things took significantly longer.

Take the configuration file changes that OpenSMTPD went through.  Which one do you break first?  Either way, there's at least some period of time where the tests are broken.

## `smtpscript` is special
What I mean is that it's a standalone utility, and isn't actually a test itself.  As long as it can be treated as a separate thing there's nothing wrong with keeping it over there.  It's pretty common for programs to import different test runners after all.

## Tests should work for both OpenBSD and Portable
It sounds like the problem is my tests don't integrate into the current testing framework.  That's my fault, but it sounds like there should at least be a mention of `smtpscript`, and how to run it in the Readme.

## How I plan to fix the issue
I'll try to re-work the tests and commits so you can push those up to master, and the testing framework will just run `smtpscript`.

poolpOrg:
ok fair enough :-)

poolpOrg:
Let's merge as is and work can be done to make a proper testing/regress workflow because today there's no such thing.

PR #914, opened 2019-03-21
WIP: add "queue directory <path>" config option

catern:
This allows the user to specify a different directory for the spool.

TODO: document in manpage

I tested this with a build and configtest, but haven't verified that everything is still actually working. Want to make sure this approach is fine first.


poolpOrg:
PR was not updated to address the comments so assuming original poster lost interest.

PR #915, opened 2019-03-26
WIP: make paths and users configurable

catern:
This patchset allows running many parts of OpenSMTPD with a non-standard configuration.

See the commit messages for more information on each commit.

This is enough to run OpenSMTPD as an unprivileged user without any system-wide configuration, along with a few other hacks: https://github.com/catern/OpenSMTPD/pull/1

### DIFF
--- a/smtpd/ca.c
+++ b/smtpd/ca.c
@@ -83,7 +83,7 @@ ca(void)
 	if ((pw = getpwnam(SMTPD_USER)) == NULL)
 		fatalx("unknown user " SMTPD_USER);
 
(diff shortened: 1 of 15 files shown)


poolpOrg:
Part of the changeset is acceptable upstream for both native and portable versions, but a large part isn't and the PR has grown into a very large diff rather than multiple small individual ones.

This can't be merged as is, if there is still interest please split into individual changes as it allows merging the bits we're ok with without having to reject the whole if there are parts that at the very least require discussions.

PR #925, opened 2019-07-22
smtpd: Fix compliance of SMTP ENHANCEDSTATUSCODE replies with RFC 2034.

stephanbosch:
Two problems:
- RFC 2034 Section 3 (item 4) requires that the status code is always followed
  by one or more spaces. However, OpenSMTPD has a colon right after the status
  code in many of the replies it sends.
- RFC 2034 Section 4 states: "When responses are continued across multiple lines
  the same status code must appear at the beginning of the text in each line of
  the response." However, OpenSMTPD only has the enhanced status code for the
  last reply line it sends. Still, there is only one case in which it sends
  such a multi-line reply: the HELP command.

I couldn't compile OpenSMTPd at my end, so this change isn't actually tested in any way. I made this change based on behavior seen by others and by inspecting the source code.

### DIFF
--- a/smtpd/smtp_session.c
+++ b/smtpd/smtp_session.c
@@ -721,7 +721,7 @@ smtp_session_imsg(struct mproc *p, struct imsg *imsg)
 			break;
 		case LKA_TEMPFAIL:
 			smtp_tx_free(s->tx);
-			smtp_reply(s, "421 %s: Temporary Error",
+			smtp_reply(s, "421 %s Temporary Error",
 			    esc_code(ESC_STATUS_TEMPFAIL, ESC_OTHER_MAIL_SYSTEM_STATUS));
 			break;
 		}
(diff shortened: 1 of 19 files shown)


poolpOrg:
Thanks, will review and merge shortly ;-)

smtpd/smtp_session.c
```diff
@@ -1833,27 +1833,28 @@ smtp_proceed_data(struct smtp_session *s, const char *args)
...
 static void
 smtp_proceed_help(struct smtp_session *s, const char *args)
 {
-	smtp_reply(s, "214- This is " SMTPD_NAME);
-	smtp_reply(s, "214- To report bugs in the implementation, "
-	    "please contact bugs@openbsd.org");
-	smtp_reply(s, "214- with full details");
-	smtp_reply(s, "214 %s: End of HELP info",
-	    esc_code(ESC_STATUS_OK, ESC_OTHER_STATUS));
+	const char *esc_code = esc_code(ESC_STATUS_OK, ESC_OTHER_STATUS);
```
greecemunky: Just thinking out loud, and on the basis it's not been compiled and tested... wouldn't there be a conflict between variable and function names being identical?
To remove any ambiguity, I suggest changing the local variable name to something like `code_esc`.

stephanbosch:
LOL, yeah, that won't compile.

Fixed.

EmperorArthur:
Hey, it's my bad for not making the documentation clear on this.

The docker build process compiles OpenSMTPD.  Since it's docker everything needed for compilation is pulled in automatically.  Other than Docker itself there are no dependencies for you to worry about!

```
docker build -t opensmptd .
```

stephanbosch:
Well, I just tried to compile it directly in my local git repo and it turns out it depends on LibreSSL, which isn't readily available in Debian Buster. So, that is where I gave up compiling it. I assume it compiles now. You can test/amend that at your end; the patch is simple enough.

Note that I don't use OpenSMTPD myself. I just got a report on some problem in the interaction between OpenSMTPD and Dovect that indicated that OpenSMTPD was misbehaving. Rather than just shifting the blame, I also created a quick merge request here. :)

poolpOrg:
@stephanbosch I have fixed the failures to build with OpenSSL 1.1.x and committed to the portable branch yesterday, so as of today if you checkout the portable branch it should be just fine on Debian Buster afaik.

I noted you don't use OpenSMTPD yourself but in case someone asks you in the future, at least now you know :-)

poolpOrg:
committed, thanks !

Issue #944, opened 2019-10-10
Build with libressl is failing


ngortheone:
Ran into this today:

`OpenSMTPD $ docker build -t opensmtpd-dev . ` 

```
...
config.status: executing libtool commands
make  all-recursive
make[1]: Entering directory '/opensmtpd'
Making all in openbsd-compat
make[2]: Entering directory '/opensmtpd/openbsd-compat'
gcc -DHAVE_CONFIG_H -I. -I..  -I../smtpd -I../openbsd-compat -I../openbsd-compat/err_h -I/usr/include   -g -O2  -fPIC -DPIC -Wall -Wpointer-arith -Wuninitialized -Wsign-compare -Wformat-security -Wsizeof-pointer-memaccess -Wno-pointer-sign -Wno-unused-result -fno-strict-aliasing -fno-builtin-memset -D_BSD_SOURCE -D_DEFAULT_SOURCE  -c -o arc4random.o arc4random.c
arc4random.c:167:21: error: macro "arc4random_stir" passed 1 arguments, but takes just 0
 arc4random_stir(void)
                     ^
arc4random.c:168:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before '{' token
 {
 ^
make[2]: Leaving directory '/opensmtpd/openbsd-compat'
make[2]: *** [Makefile:446: arc4random.o] Error 1
make[1]: Leaving directory '/opensmtpd'
make[1]: *** [Makefile:419: all-recursive] Error 1
make: *** [Makefile:351: all] Error 2
The command '/bin/sh -c ./bootstrap &&     ./configure --with-gnu-ld --sysconfdir=/etc/mail --with-path-empty=/var/lib/opensmtpd/empty/ &&     make &&     make install' returned a non-zero code: 2
```

I am not a C dev, but it looks like a conflicting name with stdlib

ngortheone:
Steps to reproduce
1. Checkout portable branch
2. Run `docker build -t opensmtpd-portable-dev .`

ngortheone:
I am completely puzzled. Same code compiles fine on debain. (compiler version is the same)

ngortheone:
So, here is the problem, but I dunno how to fix it.


./openbsd-compat/openbsd-compat.h
```c
#if defined(HAVE_ARC4RANDOM_STIR)
void arc4random_stir(void);
#elif defined(HAVE_ARC4RANDOM) || defined(LIBRESSL_VERSION_NUMBER)
[...]

PR #945, opened 2019-10-11
Swap libressl with openssl in alpine docker

ngortheone:
Reason:
Latest portable branch build fails on Alpine Linux (and probably on
other distors where libressl is available) with several distinct errors.
Build succeds with OpenSSL.

After reading @poolpOrg 's [July update](https://poolp.org/posts/2019-07-27/july-2019-report-tons-of-smtpd-work-mostly/) it became clear that why.

Fix for https://github.com/OpenSMTPD/OpenSMTPD/issues/944


ngortheone:
Also added some minor tweaks to make resulting container usable with or without mounted volumes

poolpOrg:
can you split this into individual diffs ?

there are bits that i'd like to merge but others that I'd like to discuss :-/

ngortheone:
ack

ngortheone:
@poolpOrg can you briefly describe what parts do you like so I can split more accurately?

PR #946, opened 2019-10-13
Fix broken docker build

ngortheone:
Split of  https://github.com/OpenSMTPD/OpenSMTPD/pull/945
Replace libressl with openssl

Fixes the immediate problem of broken docker build.


poolpOrg:
Thanks

PR #960, opened 2019-10-27
doc: add 6.6.0 changes and more

ngortheone:
- add changes file to reflect what is new in 6.6.0 version (as discussed in https://github.com/OpenSMTPD/OpenSMTPD/issues/944)
- all markdown headers are atx style now to allow more then 2 levels (mix of setext and atx does not work well)
- moved docker documentation to docker folder

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,31 @@
+# Release 6.6.0 #
+
+## Dependencies note:
+
+This release builds with LibreSSL > 3.0.2 or OpenSSL > 1.1.0.
+
+It's preferable to depend on LibreSSL as OpenSMTPD is written and tested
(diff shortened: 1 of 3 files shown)


poolpOrg:
I'll merge this because it is better than what we have, but I think we should not put explicit versions to our dependencies but rather put something along the lines of "unless stated otherwise explicitely, OpenSMTPD relies on the latest version of each dependency".

ngortheone:
Sure, lmk next time in code review so I can fix it. Do you want a followup PR to remove explicit versions and a disclaimer "f "unless stated otherwise explicitely, OpenSMTPD relies on the latest version of each dependency"."?

poolpOrg:
sure !

Issue #958, opened 2019-10-26
Error compiling 6.6.0p1 with LibreSSL


ArchangeGabriel:
On ArchLinux, with LibreSSL 3.0.2 and libasr 1.0.3:
```
gcc -DHAVE_CONFIG_H -I. -I..  -I../smtpd -I../openbsd-compat -I../openbsd-compat/err_h -I/usr/include -D_FORTIFY_SOURCE=2  -march=x86-64 -mtune=generic -O2 -pipe -fno-plt  -fPIC -DPIC -Wall -Wpointer-arith -Wuninitialized -Wsign-compare -Wformat-security -Wsizeof-pointer-memaccess -Wno-pointer-sign -Wno-unused-result -fno-strict-aliasing -fno-builtin-memset -I/usr/include/libressl -L/usr/lib/libressl -Wl,-rpath=/usr/lib/libressl -D_BSD_SOURCE -D_DEFAULT_SOURCE  -c -o arc4random.o arc4random.c
arc4random.c:167:21: error: macro "arc4random_stir" passed 1 arguments, but takes just 0
  167 | arc4random_stir(void)
      |                     ^
In file included from includes.h:71,
                 from arc4random.c:27:
openbsd-compat.h:130: note: macro "arc4random_stir" defined here
  130 | # define arc4random_stir()
      | 
arc4random.c:168:1: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘{’ token
  168 | {
      | ^
```

ngortheone:
@ArchangeGabriel 

I have seen this here:
https://github.com/OpenSMTPD/OpenSMTPD/issues/944


> So, here is the problem, but I dunno how to fix it.
> 
> ./openbsd-compat/openbsd-compat.h
> 
> ```c
> #if defined(HAVE_ARC4RANDOM_STIR)
> void arc4random_stir(void);
> #elif defined(HAVE_ARC4RANDOM) || defined(LIBRESSL_VERSION_NUMBER)
> /* Recent system/libressl implementation; no need for explicit stir */
> # define arc4random_stir()
> #else
> /* openbsd-compat/arc4random.c provides arc4random_stir() */
> void arc4random_stir(void);
> #endif
> ```
> 
> ./openbsd-compat/arc4random.c
> 
> ```c
> void
> arc4random_stir(void)
> {
> 	_ARC4_LOCK();
> 	_rs_stir();
[...]

PR #961, opened 2019-10-27
docker: archlinux build

ngortheone:
Archlinux build with glibc, libressl, gcc

### DIFF
--- a/docker/Dockerfile.archlinux
+++ b/docker/Dockerfile.archlinux
@@ -1,0 +1,43 @@
+FROM archlinux
+
+WORKDIR /opensmtpd
+
+RUN pacman -Suy --noconfirm \
+  make \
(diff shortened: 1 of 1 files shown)


ArchangeGabriel:
You’re not building with LibreSSL that way, but with OpenSSL. Hence why it works.

ngortheone:
@ArchangeGabriel 
I am a bit confused. (I am also not a C expert)

> You’re not building with LibreSSL that way, but with OpenSSL. Hence why it works.
There is no openssl in dockerfile, 

Can you elaborate?

ArchangeGabriel:
By default you will only find the OpenSSL headers/pkgconfig files. To have the LibreSSL ones, you need to manually specify them as told in #958.

PR #970, opened 2019-11-02
ci: coverity scan workflow

ngortheone:
Adds Coverity scan scripts:
- shell script to run locally
- workflow to run everyday on GitHub Actions

And some minor structure changes:
- docker workflow split into multiple workflows, one per Dockerfile.
- ubuntu build workflow converted to Dockerfile for consistency.
- created top level `ci` dir
- `docker` dir moved into `ci` dir
- created `scripts` dir inside `ci` dir

@poolpOrg I know that you wanted to run coverity off a separate branch. Unfortunately it is not as easy to do as just trigger a cron action (cron action only works with default branch). 
I will fix it if it gets too nosiy with failures:
I will have to write a custom action for us to push to a branch on cron and then create "on push" workflow for scan branch.

P.S. Coverity token is not leaked in the output :) (`set +x` hides command invocations in logs)


poolpOrg:
It's ok if ran from a cron, the reason I didn't want it on the portable branch was that I didn't want it to be triggered by commits and merges to that branch due to the scan.coverity.com quotas, i like your pull request.

PR #998, opened 2019-11-18
Fix typos in smtpd.conf.5

ryanakca:
The word "criteria" is already the plural of "criterion".


ngortheone:
@poolpOrg 
Failed archlinux does not look related to the change. It looks like old docker image that is missing librarchive
https://www.archlinux.org/news/required-update-to-recent-libarchive/

ngortheone:
looks like it was fixed, I just re-ran the test

poolpOrg:
committed to OpenBSD and synchronized on github, thanks

Issue #994, opened 2019-11-12
OpenSMTPD terminating lmtp session after LHLO


michele:
Hello,

I've been banging my head against this issue since yesterday night: basically I've got smtpd configured to deliver email to Dovecot through LMTP (tried both as a socket and ip:port), however, it fails to do so and it doesn't through any errors in the logs.
So I've activate LMTP raw logs in Dovecot and this is what I get:

IN:

```                                                                                                                                                   
1573560929.960497 LHLO localhost
```

OUT:

```
1573560929.960425 220 mail.example.com Dovecot ready.
1573560929.960545 250-mail.example.com
1573560929.960545 250-8BITMIME
1573560929.960545 250-CHUNKING
1573560929.960545 250-ENHANCEDSTATUSCODES
1573560929.960545 250-PIPELINING
1573560929.960545 250 STARTTLS
```

It seems like it's terminating the connection early. I see there was an old bug about this but it's supposed to be fixed.

I'm running Dovecot 2.3.7.2 and OpenSMTPD 6.6.0-portable on Alpine.

Any help would be highly appreciated. Thanks!

michele:
I tried upgrading to Dovecot 2.3.8 and OpenSMTPD 6.6.1 but things haven't changed

whataboutpereira:
What's your OpenSMTPD config concerning Dovecot?

Did you try running smtpd -dv?

michele:
@whataboutpereira I'm pasting both below.

OpenSMTPD config:

```
action "virtact" lmtp "/sockets/lmtp.sock" virtual <virtuals>                                                                                                                                                                                                                                                          
action "local_mail" lmtp "/sockets/lmtp.sock" alias <aliases>                                                                                                                               
                                                                                                          
match from any for domain <domains> action "virtact"                                                                                                                                        
match for local action "local_mail"                                                                                                                                                         
```

I'm running smtp with -dv, but there's not much in the logs, here's a sample:

```
7f6c28bbac82816e smtp connected address=64.147.123.24 host=wout1-smtp.messagingengine.com                                                                                                   
debug: looking up pki "mail.example.com"                                                                                                                                                     
debug: session_start_ssl: switching to SSL                                                                                                                                                  
debug: pony: rsae_priv_enc                                                                                                                                                                  
7f6c28bbac82816e smtp tls ciphers=TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256                                                                                                                   
[...]

PR #999, opened 2019-11-22
workaround for musl's implementation of fflush() in mail.lmtp

fgma:
I've added a workaround for the problem discussed in https://github.com/OpenSMTPD/OpenSMTPD/issues/994. Right now it is only tested on alpine linux.

### DIFF
--- a/smtpd/mail.lmtp.c
+++ b/smtpd/mail.lmtp.c
@@ -48,15 +48,15 @@ struct session {
 	int		n_rcpts;
 };
 
-static FILE *lmtp_connect(const char *);
-static void lmtp_engine(FILE *, struct session *);
+static int lmtp_connect(const char *);
+static void lmtp_engine(int, struct session *);
 static void stream_file(FILE *);
 	
 int
 main(int argc, char *argv[])
 {
 	int ch;
-	FILE *conn;
+	int conn;
 	const char *destination = "localhost";
 	struct session	session;
 
@@ -99,7 +99,7 @@ main(int argc, char *argv[])
 	return (0);
 }
 
-static FILE *
+static int
 lmtp_connect_inet(const char *destination)
 {
 	struct addrinfo hints, *res, *res0;
@@ -173,10 +173,10 @@ lmtp_connect_inet(const char *destination)
 		errx(EX_TEMPFAIL, "%s", cause);
 
 	free(destcopy);
-	return fdopen(s, "r+");
+	return s;
 }
 
-static FILE *
+static int
 lmtp_connect_unix(const char *destination)
 {
 	struct sockaddr_un addr;
@@ -197,10 +197,10 @@ lmtp_connect_unix(const char *destination)
 	if (connect(s, (struct sockaddr *)&addr, sizeof addr) == -1)
 		err(EX_TEMPFAIL, "connect");
 
-	return fdopen(s, "r+");
+	return s;
 }
 
-static FILE *
+static int
 lmtp_connect(const char *destination)
 {
 	if (destination[0] == '/')
@@ -209,17 +209,42 @@ lmtp_connect(const char *destination)
 }
 
 static void
-lmtp_engine(FILE *conn, struct session *session)
+lmtp_engine(int fd_read, struct session *session)
 {
+	int fd_write = 0;
+	FILE *file_read = 0;
+	FILE *file_write = 0;
 	char *line = NULL;
 	size_t linesize = 0;
 	ssize_t linelen;
 	enum phase phase = PHASE_BANNER;
 
+	fd_write = dup(fd_read);
+
+	if (fd_write == -1){
+		err(EX_TEMPFAIL, "dup");
+		return;
+	}
+
+	file_read = fdopen(fd_read, "r");
+
+	if (!file_read){
+		err(EX_TEMPFAIL, "fdopen");
+		return;
+	}
+
+	file_write = fdopen(fd_write, "w");
+
+	if (!file_write){
+		err(EX_TEMPFAIL, "fdopen");
+		return;
+	}
+
 	do {
-		fflush(conn);
-		if ((linelen = getline(&line, &linesize, conn)) == -1) {
-			if (ferror(conn))
+		fflush(file_write);
+
+		if ((linelen = getline(&line, &linesize, file_read)) == -1) {
+			if (ferror(file_read))
 				err(EX_TEMPFAIL, "getline");
 			else
 				errx(EX_TEMPFAIL, "unexpected EOF from LMTP server");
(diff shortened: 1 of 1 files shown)


smtpd/mail.lmtp.c
```diff
@@ -209,17 +209,42 @@ lmtp_connect(const char *destination)
...
+	FILE *file_read = 0;
+	FILE *file_write = 0;
 	char *line = NULL;
 	size_t linesize = 0;
 	ssize_t linelen;
 	enum phase phase = PHASE_BANNER;
 
+	fd_write = dup(fd_read);
+
+	if (fd_write == -1){
```
poolpOrg: coding style:

if ((fd_write = dup(fd_read)) == -1) {
    ....
}

smtpd/mail.lmtp.c
```diff
@@ -209,17 +209,42 @@ lmtp_connect(const char *destination)
...
+		err(EX_TEMPFAIL, "dup");
+		return;
```
poolpOrg: err() doesn't return, so this can be replaced with a single:

if ((fd_write = dup(fd_read)) == -1)
    err(EX_TEMPFAIL, "dup"):

smtpd/mail.lmtp.c
```diff
@@ -209,17 +209,42 @@ lmtp_connect(const char *destination)
...
+		err(EX_TEMPFAIL, "fdopen");
+		return;
+	}
+
+	file_write = fdopen(fd_write, "w");
+
+	if (!file_write){
+		err(EX_TEMPFAIL, "fdopen");
+		return;
+	}
```
poolpOrg: same as comments above

fgma:
I did some cleanup.

smtpd/mail.lmtp.c
fgma: Done

ngortheone:
@fgma can you briefly describe for non-C people how this fix works? thanks

fgma:
If you open a file/socket/etc. you get back a file handle. You can use it for reading or writing bytes to it. fdopen() adds a layer on top of the raw file handle. With this additional layer you can use stream functions like getline() and fprintf() for a more high level access to the file handle.

In the original code fflush() is used to ensure that a lmtp command is sent before waiting for the reply to it. OpenBSD and glibc will flush just the write buffer when calling fflush(). Musl additionally flushes the read buffer. When the lmtp server replies with a multiline reply only one line is read per loop using getline() afterwards the rest of the reply gets flushed by fflush().

To get around that I added a call to dup() which duplicates the file handle. This new file handle references the same file/socket/etc. Therefor I can fdopen() both of them and use one exclusively for reading and the other one exclusively for writing. Therefor fflush() will only flush data in the high level layer for one direction of data flow.

smtpd/mail.lmtp.c
```diff
@@ -209,17 +209,30 @@ lmtp_connect(const char *destination)
...
+	FILE *file_write = 0;
 	char *line = NULL;
 	size_t linesize = 0;
 	ssize_t linelen;
 	enum phase phase = PHASE_BANNER;
 
+	if ((fd_write = dup(fd_read)) == -1)
+		err(EX_TEMPFAIL, "dup");
+
+	if (!(file_read = fdopen(fd_read, "r")))
```
poolpOrg: OpenBSD style checks pointers against NULL, not as booleans

smtpd/mail.lmtp.c
```diff
@@ -209,17 +209,30 @@ lmtp_connect(const char *destination)
...
+		err(EX_TEMPFAIL, "fdopen");
+
+	if (!(file_write = fdopen(fd_write, "w")))
```
poolpOrg: same

poolpOrg:
almost ok w/ me, two minor changes and we're good

smtpd/mail.lmtp.c
poolpOrg: just to clarify, I'm refering to the fdopen() call, it should read:

if ((file_read = fdopen(fd_read, "r")) == NULL)

fgma: Fixed in https://github.com/OpenSMTPD/OpenSMTPD/pull/999/commits/0d9aebb4197cf99020afed950fd9f43afe38427e

smtpd/mail.lmtp.c
fgma: Fixed in https://github.com/OpenSMTPD/OpenSMTPD/pull/999/commits/0d9aebb4197cf99020afed950fd9f43afe38427e

poolpOrg:
perfect :+1:

I will leave a bit of time for @snimmagadda to comment as he wrote the LMTP code and I'm not an LMTP user, but if it takes too long I'll commit.

Thanks !

fgma:
@poolpOrg Thanks, when are you going to make a new release which integrates this PR?

ngortheone:
> If you open a file/socket/etc. you get back a file handle. You can use it for reading or writing bytes to it. fdopen() adds a layer on top of the raw file handle. With this additional layer you can use stream functions like getline() and fprintf() for a more high level access to the file handle.
> 
> In the original code fflush() is used to ensure that a lmtp command is sent before waiting for the reply to it. OpenBSD and glibc will flush just the write buffer when calling fflush(). Musl additionally flushes the read buffer. When the lmtp server replies with a multiline reply only one line is read per loop using getline() afterwards the rest of the reply gets flushed by fflush().
> 
> To get around that I added a call to dup() which duplicates the file handle. This new file handle references the same file/socket/etc. Therefor I can fdopen() both of them and use one exclusively for reading and the other one exclusively for writing. Therefor fflush() will only flush data in the high level layer for one direction of data flow.

Hey @poolpOrg, over time memory of such workarounds get blurry or lost and it makes it hard to understand for newcomers why certain hacks/fixes/workarounds were made. @fgma 's explanation  looks like a valuable info for new people who want to dive into the codebase.
In other projects people leave this kind of explanations as a comments in the code and I think we should do that too. What do you think?

poolpOrg:
@fgma I'll commit this today and the work-around will be part of the next minor release in December, very likely around the 9th since that's when my next week of opensource begins.

poolpOrg:
@ngortheone you meant on github or in the actual code ?

I'm ok adding more comments to the code when it unblurs a complex logic but there needs to be a balance because too much info hinders readability.

In this case, I don't think it is necessary to add a comment because the code is pretty obvious with a read_file stream used for reads and a write_file stream used for writes, I doubt any C developer will get confused by that. The musl backstory isn't so much interesting here because the code from this diff makes sense disregarding the musl flush strategy, I would have okayed it on OpenBSD and glibc systems without knowing about musl.

PR #1000, opened 2019-11-24
Document that the local socket can also be filtered

ryanakca:
This fact is useful to people whose MUA sends mail using the sendmail
command, and who want to pass this mail through filters.


poolpOrg:
awwww you also changed the wording for other options, let me check with our man people if they agree with this change because the wording gets reviewed too.

poolpOrg:
I have committed a slightly altered version of your diff so mask-src wording is identical to the one in the listen on interface case, credited you in the commit log, thanks for your contrib.

PR #1006, opened 2019-11-25
ci: openbsd build

ngortheone:


ngortheone:
@poolpOrg this one is ready, but need to rerun all checks after #1007 is mergerd

Issue #1012, opened 2019-11-30
OpenSMTPD does not pick up offline mail on Linux


patrakov:
Test case:

1. Install OpenSMTPD from Linux packages. I have tried Debian Unstable and Arch Linux (both have OpenSMTPD 6.6.1p1).
2. Start it, because at least Arch Linux relies on it to create the hierarchy in /var/spool/smtpd
3. Stop it. I.e. pretend that it was stopped due to some accident.
4. Create an email message file, test.txt (with headers, blank line, and body)
5. Attempt to send it: sendmail -f you@yourmail.com -t < test.txt
6. The sendmail wrapper (around smtpctl) puts the message into /var/spool/smtpd/offline and returns 75, as expected.
7. Start smtpd. You can even add some debugging: smtpd -d -v -T all

Result:

```
...
debug: smtpd: scanning offline queue...
debug: smtpd: offline scanning done
...
```

... and nothing logged in between, no mail sent, and the file still in the queue.

This works (sends the email) on OpenBSD.

Some debugging with cozy[m] on IRC revealed that this is due to the different resulting permissions on the queued message:

OpenBSD: the group is inherited from the /var/spool/smtpd/offline directory:
```
-rw-------  1 root  _smtpq  238 Nov 30 23:54 1575140075.UdvOCF8AZr
```

Linux: the group is not inherited:

```
-rw------- 1 root root 242 Nov 30 19:56 1575143789.XXXXyiq27C
```

This is explained by different directory group ownership semantics on Linux and OpenBSD, see http://perfec.to/shared/shared.html#heading2

OpenSMTPD has this code that checks the `stat` results: https://github.com/OpenSMTPD/OpenSMTPD/blob/c139eb1610e931739d6cde4194c9560124b08165/smtpd/smtpd.c#L1604

So in effect, it relies on the BSD group semantics (that files in a directory inherit the group from the directory by default), which does not hold on Linux by default.

To fix this, smtpd on Linux should create the /var/spool/smtpd/offline directory with the setgid bit set, in order to force BSD semantics.

patrakov:
Another important difference is that permissions of /usr/sbin/smtpctl differ.

```
# Debian:
-rwxr-xr-x 1 root root 211896 Nov 19 17:06 /usr/sbin/smtpctl
# Arch:
-r-xr-sr-x 1 root root 163440 Nov  6 16:33 /usr/sbin/smtpctl
# OpenBSD:
[...]

PR #1013, opened 2019-12-01
smtpctl: check correct egid at startup

patrakov:
Too many Linux distributions (Debian, Ubuntu, Arch, Fedora) got smtpctl
permissions wrong. The typical pattern (applies to Arch) is:

 * The build machine does not have the smtpq group
 * The build runs as non-root with --with-group-queue=smtpq
 * The smtpq group is created in post-install via systemd-sysusers
 * There is nothing that makes smtpctl setgid smtpq
 * The package even seems to work until offline delivery is attempted

Let's make sure users complain to the distributors if the above happens,
instead of having a silently broken installation.

See also the (invalid) ticket #1012.

### DIFF
--- a/smtpd/smtpctl.c
+++ b/smtpd/smtpctl.c
@@ -1061,13 +1061,23 @@ do_spf_walk(int argc, struct parameter *argv)
 int
 main(int argc, char **argv)
 {
-	gid_t		 gid;
+	gid_t		 egid, gid;
+	struct group    *gr;
 	int		 privileged;
 	char		*argv_mailq[] = { "show", "queue", NULL };
 
 #ifndef HAVE___PROGNAME
 	__progname = ssh_get_progname(argv[0]);
 #endif
+
+	/* Sanity check that too many Linux distros fail */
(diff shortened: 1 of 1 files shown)


poolpOrg:
I have made slight changes to your diff, it reads fine to me but I wonder if we should not turn these warnx() into errx(). This would not prevent the daemon from running since the errx() would trigger in `smtpctl` but it would prevent `smtpctl` from running any further when it is obviously broken if running with wrong privileges.

Your opinion ?

patrakov:
For the installed copy of smtpctl errx() obviously makes sense. However, I was afraid to break your test procedures, that's why opted for warnx(). I don't have a strong opinion.

poolpOrg:
Then I'll switch to errx() and merge :-)

poolpOrg:
Thanks !

ngortheone:
@ArchangeGabriel @jirutka @ryanakca 
Guys, this looks like something that has to be taken care of downstream.

PR #1018, opened 2019-12-31
openbsd-compat: fix out-of-tree builds

Redfoxymoon:
In out-of-tree builds anything inside openbsd-compat/ including local files prefixed with smtpd/ are not found, including the source base directory takes care of that.


ngortheone:
@Redfoxymoon thanks for your PR.

This looks fairly innocent, but I'd still wait for @poolpOrg to vet this.

PR #1019, opened 2019-12-31
ci: ansible testing framework for freebsd build + alpine 3.11 on libressl

ngortheone:
Two things are going on here:
- FreeBSD build on SourceHut is now using my new ansible framework. It covers very small area at the moment - build and simple mail send to a local recipient, but it is easily extendible to include more tests or more platforms. I plan to include all linux builds in the nearest future.
- Alpine container upgraded to 3.11 (latest release) and switched to libressl

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -13,8 +13,11 @@ packages:
 - python37
 - python-3.7_3,2
 sources:
+- https://github.com/OpenSMTPD/OpenSMTPD
 - https://github.com/OpenSMTPD/ci
 tasks:
 - ansible: |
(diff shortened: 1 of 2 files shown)


ngortheone:
Don't merge just yet. I need to fix git ref passing from sourcehut

ngortheone:
This is now ready to be reviewed.
Note that new ansible testing framework that FreeBSD build uses is located at https://github.com/OpenSMTPD/ci

ngortheone:
It is probably best to squash this PR - don't want to have all my "work commits" to pollute portable history

PR #1049, opened 2020-04-14
Document junking in smtpd.conf(5)

ryanakca:
The man page does not explain what the `junk` decision means when filtering or what happens if you do not have a `Junk` folder. This diff documents this.


poolpOrg:
no worries, it's ok but I was waiting for @ericfaurot to commit to OpenBSD for merging this otherwise we risk having different documentation in OpenSMTPD 6.7.0 for OpenBSD and for other systems

PR #1055, opened 2020-04-24
Use correct libexec path in portable man pages

ryanakca:
smtpd.conf(5) in the portable branch refers to the directory `/usr/local/libexec/smtpd/`. This may not be the correct location on non-OpenBSD systems (it is not on Debian). Update the Makefile to update this location at compile time.


ngortheone:
It looks like it is time to update freebsd build:
`pkg: No packages available to install matching 'py37-ansible-2.8.7' have been found in the repositories`

ngortheone:
@ryanakca 
it looks like this change adjusts the path to libexec according to current `--prefix`
I am not an autotools expert, but I have compiled man pages before and after and it was the only change that I noticed

ngortheone:
@poolpOrg I am merging this one

poolpOrg:
ok !

PR #1063, opened 2020-05-23
compat: pipe2: enable O_CLOEXEC correctly and support O_NONBLOCK

zx2c4:
O_CLOEXEC must be enabled with F_SETFD, not F_SETFL. Prior passing
FD_CLOEXEC to F_SETFL resulted in wrong behavior. While we're fixing
this, we can also add support for O_CLOEXEC, which we enable via
F_SETFL.

### DIFF
--- a/openbsd-compat/pipe2.c
+++ b/openbsd-compat/pipe2.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020 Gilles Chehade <gilles@poolp.org>
+ * Copyright (C) 2020 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
  *
  * Permission to use, copy, modify, and distribute this software for any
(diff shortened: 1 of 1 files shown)


poolpOrg:
Nope, the flags to pipe2() are not the same as fcntl() ones, you must recreate the flags, something along the lines of:

    int fflags;
    if (flags & O_CLOEXEC)
        fflags |= FD_CLOEXEC

only O_CLOEXEC and O_NONBLOCK are needed

zx2c4:
Thanks for pointing this out. I took a closer look, and actually your original code was wrong. FD_CLOEXEC needs F_SETFD, not F_SETFL, according to manpages and fs/fcntl.c.

I'll send a fix accordingly.

zx2c4:
New commit pushed.

poolpOrg:
Thanks

Issue #1069, opened 2020-06-03
do_encrypt adds extra argument and causes encrypt to fail


Kimplul:
If I try calling 
```smtpctl encrypt *any string*``` 
I get the error
 ```usage: encrypt <string>```

I dug around and found out that ```do_encrypt``` in ```usr.sbin/smtpd/smtpctrl.c``` calls ```encrypt``` with 
```execl(PATH_ENCRYPT, "encrypt", "--", p, (char *)NULL);```
 but ```contrib/libexec/encrypt.c``` only takes one argument, and adding ```--``` as an argument in front of the string to be encrypted causes the error.

Is there a reason for why the two dashes were added and should ```encrypt.c``` or ```smptctrl.c``` be changed?

**Comment by studentiks** [comment]
Doing `echo -n "my_password" | smtpctl encrypt` outputs hash of "--" instead of "my_password". Workaround (in archlinux) is to use `echo -n "my_password" | /usr/lib/smtpd/opensmtpd/encrypt`

**Comment by ssartor** [comment]
Does not look like this fix was included in the latest portable version (6.7.1p1)

**Comment by gearnode** [comment]
While the patch was release, on FreeBSD you can use:
```
echo "mypassword" | /usr/local/libexec/opensmtpd/encrypt
```

poolpOrg:
The fix was merged and will be part of the upcoming 6.8.0p1 release

PR #1073, opened 2020-07-02
fix bug with "smtpctl encrypt"

xukl:
encrypt.c will exit if `argc > 2`. However, the code at smtpctl.c:L969 executes `encrypt` with wrong number of arguments when `p != NULL`, which leads to a bug that the program complains `usage: encrypt <string>` when we run `smtpctl encrypt some_string`.
The code  `"--", ` was involved in commit `4c624cc`. Before that commit, the code had always been `execl(PATH_ENCRYPT, "encrypt", p, (char *)NULL);`.

### DIFF
--- a/usr.sbin/smtpd/smtpctl.c
+++ b/usr.sbin/smtpd/smtpctl.c
@@ -966,7 +966,7 @@ do_encrypt(int argc, struct parameter *argv)
 
 	if (argv)
 		p = argv[0].u.u_str;
-	execl(PATH_ENCRYPT, "encrypt", "--", p, (char *)NULL);
+	execl(PATH_ENCRYPT, "encrypt", p, (char *)NULL);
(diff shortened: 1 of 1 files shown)


ArchangeGabriel:
I’ve been fighting for 2 h debugging SMTP because of this. `echo passwd | smtpctl encrypt` does not work because of this, while the supposed interactive mode `smtpctl encrypt` returns directly an encrypted chain (I suppose for `--`…).

xukl:
I found that @Kimplul had found this bug as early as 3 Jun and open the issue #1069. I'm pretty sure this PR can fix this bug. So... can anyone merge this PR? Or is there anything I still need to do?

poolpOrg:
needed to understand why this was introduced and didn't have time to investigate earlier.

fixed

PR #1093, opened 2020-10-15
Fix out-of-tree build

DemiMarie:
Automake treats $(top_srcdir) specially, so a variable that expands to
it won't work.  There were also bugs in the Makefile.am files that broke
`make distcheck`


poolpOrg:
Thanks

Issue #1059, opened 2020-05-05
Broken syslog messages from OpenSMTPD v6.7.0 on CentOS 8.


whataboutpereira:
I tried out current portable a few days ago on CentOS 8, building it the exact same way that my 6.6.4p1 was built previously.

I'm running into weird log entries in /var/log/maillog:

```
May  2 12:32:13 centos8 smtpd[124028]: info: OpenSMTPD 6.7.0-portable starting
May  2 12:32:50 centos8 journal[124033]: y express[124033]: 6ad20154b7c9470a smtp connected address=local host=centos8

                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Syslog seems to get confused about the process that's logging.

I'm running it with the same systemd unit script as 6.6.4p1 (posted here: https://github.com/OpenSMTPD/OpenSMTPD/issues/1039#issuecomment-623426652).

Any ideas?

dfateyev:
I can confirm it not only for EL7 (`systemd-239`), but all recent systemd versions, e.g. `systemd-245` under recent Fedora:

```
May 17 04:17:43 fedora32-amd64.svc.xxxx.xxx smtpd[19980]: y express[19980]: eebcd24b54b9d4a3 smtp connected address=local host=fedora32-amd64.svc.xxxx.xxx
May 17 04:17:43 fedora32-amd64.svc.xxxx.xxx smtpd[19980]: y express[19980]: eebcd24b54b9d4a3 smtp message msgid=a9f5e0cc size=529 nrcpt=1 proto=ESMTP
May 17 04:17:43 fedora32-amd64.svc.xxxx.xxx smtpd[19980]: y express[19980]: eebcd24b54b9d4a3 smtp envelope evpid=a9f5e0cc7f3ac069 from=<root@fedora32-amd64.svc.xxxx.xxx> to=<xxx@xxx.com>
May 17 04:17:43 fedora32-amd64.svc.xxxx.xxx smtpd[19980]: y express[19980]: eebcd24b54b9d4a3 smtp disconnected reason=quit
May 17 04:17:43 fedora32-amd64.svc.xxxx.xxx smtpd[19980]: y express[19980]: eebcd24f3e08857b mta connecting address=smtp://64.233.164.27:25 host=lf-in-f27.1e100.net
[...]

PR #1094, opened 2020-10-16
Fix get_progname() issues.

whataboutpereira:
Fix for https://github.com/OpenSMTPD/OpenSMTPD/issues/1059

Apparently libc __progname doesn't give the same result as get_progname() from openbsd-compat.


poolpOrg:
Hi,

This isn't the correct fix, it will not check conditionally and may break build elsewhere.

I'll be working on a fix shortly before the 6.8.0p1 release is issued.

whataboutpereira:
It fixed the naming discrepancy between ssh_get_progname and get_progname, but I guess the question is how to check for __progname that is incompatible with openbsd-compat get_progname.

poolpOrg:
Looked into it as I'm preparing the portable release, but this isn't the proper fix.

I will suggest a different fix in the coming days.

poolpOrg:
issue understood, crafting proper diff

poolpOrg:
closing this pull request as it's not correct and I committed a different fix we can discuss in the original issue

PR #1098, opened 2020-11-21
Fix macos build

gearnode:
I've edit smtpd.c maybe I need to open a contrib on OpenBSD for this one?
I can create a dedicated pull request for the source hub ci patch too.


poolpOrg:
Awwwww, thanks but can you please submit the PR without all the whitespace fixes as they are not portable specific they should be sumitted to OpenBSD separately and trickle down here when I sync the portable repository with upstream

gearnode:
@poolpOrg done :tada:

PR #1100, opened 2020-11-27
Update MacOS instructions

ryanakca:
Updated the instructions for MacOS X and Big Sur.

Maybe suggest with `--sysconfdir=/etc/mail` ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,12 @@ libasr directory:
 
     ./configure --with-libasr=/usr/local
 
-### Mac OS X:
+### MacOS:
 
-    ./configure --with-libevent=/opt/local --with-libasr=/opt/local
+    ./configure --with-libevent=/opt/local
 
+Though MacOS includes a copy of bison in the bases system, you will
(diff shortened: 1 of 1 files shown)


poolpOrg:
I don't know MacOS enough to know if there's a proper place for third-party sysconfdir, it looks like stuff is installed below /usr/local from homebrew.

Let's merge this for now and do some research before adding the sysconfdir in the README

ryanakca:
Aha, it looks like the above patch is only partially correct: it assumes the user is using MacPorts, which places everything under `/opt/local`, while Homebrew puts everything under `/usr/local`. I've never used Homebrew, but I think the corresponding invocation should be:
```
./configure --with-libevent=/usr/local
```

poolpOrg:
Yes, I use Homebrew not macports and didn't need `--with-libevent` so I guess suggesting `/usr/local` by default is the nicest approach.

I don't have strong opinions about this, the main reason I care about MacOS myself is that I can use vscode and docker to test portability but I'm not sure threre are many MacOS users out there :-p

PR #1102, opened 2020-12-09
Don't mention OpenBSD interface 'all' in portable smtpd.conf

ryanakca:
OpenBSD has an "all" interface group that includes all interfaces
(ifconfig(8)), but this interface group does not exist on other systems.
As a result, it doesn't make sense to discuss listening on this
interface in the portable version of OpenSMTPD. Instead, we should
suggest that users listen on the meta-addresses 0.0.0.0 and :: if they
want to receive mail on all interfaces.

Closes: https://bugs.debian.org/963252

### DIFF
--- a/usr.sbin/smtpd/smtpd.conf
+++ b/usr.sbin/smtpd/smtpd.conf
@@ -5,8 +5,11 @@
 
 table aliases file:/etc/mail/aliases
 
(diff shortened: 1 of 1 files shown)


ngortheone:
I support this. `all` interface always tripped me, this is especially confusing because this is in official docs, not some random blog post.

PR #1109, opened 2020-12-18
Do not hardcode the paths to cat and zcat at compile time

ryanakca:
Paths to cat and zcat are ostensibly hardcoded at compile time to ensure
that the system-wide binaries are used. Because these utilities are used
in a privileged setting to display message envelopes and contents, this
mitigates PATH-based attacks.

Unfortunately, AC_PATH_PROG sets the path to these binaries based on the
environment's PATH when configure is run, and the paths to (z)cat may
differ relative to the system on which smtpctl is run. In particular, it
may result in different paths to (z)cat on systems with and without a
merged /usr.

Instead, we should use execlp to look up the path to cat and zcat at
runtime based on the default system search path, which is set in
paths.h. This improves the portability of binaries while still
mitigating PATH-based attacks.


nckx:
The paths are still hard-coded, but there are  more of them :stuck_out_tongue: 

I prefer the current predictable behaviour over added run-time cleverness.  Is it possible to still support both?

ngortheone:
FreeBSD build fails is noise.

> pkg: No packages available to install matching 'db6' have been found in the repositories

Looks like either package name changed or something else happened. I need to dive in and look into this closer.

ngortheone:
There is no `db6` in packages. At least not anymore. I wonder how it worked before..
```
db-1.14,2                      Manipulate db(3)'s btree(3) and hash(3) databases - NetBSD port
db18-18.1.40                   Oracle Berkeley DB, Release 18.1
db5-5.3.28_7                   Oracle Berkeley DB, revision 5.3

```

ngortheone:
It was removed recently

https://www.freshports.org/databases/db6

> Deprecated DEPRECATED: latest 6.2.x version and newer not downloadable without registration
Expired This port expired on: 2020-12-02

ngortheone:
@ryanakca can you rebase to HEAD? This should fix FreeBSD CI build

PopeNobody:
Has anybody used zlib?  It's not hard, and if you link gzip into your executable, you always know where to find it!

PR #1117, opened 2020-12-26
Add libxcrypt support

breard-r:
Some systems uses [libxcrypt](https://github.com/besser82/libxcrypt/) as a replacement from their libc `crypt()` implementation. However, in order to leverage libxcrypt's features (for example, the yescrypt hashing algorithm) some changes are required. This PR changes encrypt's behavior so it can use libxcrypt's `crypt_gensalt()` if available as well as some additional strong password hashing algorithms.

There is no change on systems that does not have libxcrypt.


poolpOrg:
Thanks !

PR #1119, opened 2020-12-29
Fix broken FBSD CI build

ngortheone:
db6 package was removed from ports tree due to it is no longer
downloadable without registration. It is recommended to use db5 instead.

https://www.freshports.org/databases/db6


ngortheone:
This should unbreak https://github.com/OpenSMTPD/OpenSMTPD/pull/1109

PR #1133, opened 2021-06-24
status was set incorrectly when an envelope expired

erik-ou:


erik-ou:
surprised nobody noticed this before.  it sure makes for some confusing messages:

    A message is delayed for more than 0 seconds for the following
    list of recipients:

bo@gus.test: Envelope expired

poolpOrg:
this was merged upstream a while back

PR #1150, opened 2021-10-08
Remove check inhibiting parameters on NOOP.

### LINKED ISSUE #1148: NOOP doesn't ignore the parameter string (RFC non-conformance)
RFC 5321 contains the following text about NOOP:

> 4.1.1.9.  NOOP (NOOP)
> 
>    If a parameter string is specified, servers SHOULD ignore it.

OpenSMTPD 6.8.0p2 responds as follows:

```
 % nc 127.0.0.1 25
220 tucson ESMTP OpenSMTPD
noop
250 2.0.0 Ok
noop abc
500 5.5.4 Invalid command arguments: command does not accept arguments.
```

The key word SHOULD is defined as follows in RFC 2119:

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
> may exist valid reasons in particular circumstances to ignore a
> particular item, but the full implications must be understood and
> carefully weighed before choosing a different course.

It seems to me that the case for choosing a different course is most likely not given with the parameter string to NOOP. And therefore I believe that its implementation is non-conforming and am making the case to modify it. If this idea meets with approval, I would be happy to work on this and submit a pull request.

sjbronner:
Fix #1148

### DIFF
--- a/usr.sbin/smtpd/smtp_session.c
+++ b/usr.sbin/smtpd/smtp_session.c
@@ -1,4 +1,4 @@
-/*	$OpenBSD: smtp_session.c,v 1.434 2023/05/31 16:51:46 op Exp $	*/
+/*	$OpenBSD: smtp_session.c,v 1.435 2023/06/23 18:32:28 op Exp $	*/
 
 /*
  * Copyright (c) 2008 Gilles Chehade <gilles@poolp.org>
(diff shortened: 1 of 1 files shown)


omar-polo:
Sorry for the delay.

I've rebased and tweaked the diff to provide `smtp_check_noop()` nevertheless as leaving a NULL pointer around in a place where it could be called is a footgun.

In the [corresponding tech@ thread](https://marc.info/?l=openbsd-tech&m=168752274723303&w=2) @poolpOrg suggested that it could be benefit to allow filters to receive NOOPs too.

Thanks!

PR #1153, opened 2021-10-16
add getsubopt to openbsd-compat

satan1st:
The OpenBSD version of getsubopt has a non portable extention which is used in usr.sbin/smtpd/smtpc.c

Not sure if the configure check should also test for the extention.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -396,6 +396,10 @@ AC_CHECK_DECLS([LOG_PERROR], [], [], [
 #include <syslog.h>
 ])
 
+AC_CHECK_DECLS([suboptarg], [], [], [
+#include <stdlib.h>
+])
+
 
 #
 # CHECKS FOR COMPILER CHARACTERISTICS
@@ -2052,6 +2056,7 @@ AM_CONDITIONAL([NEED_ARC4RANDOM], [test "x$ac_cv_func_arc4random" != "xyes" -a "
 AM_CONDITIONAL([NEED_SSL_CTX_USE_CERTIFICATE_CHAIN_MEM], [test "x$ac_cv_have_decl_LIBRESSL_VERSION_NUMBER" != "xyes"])
 
 AM_CONDITIONAL([NEED_PROGNAME], [test "x$ac_cv_libc_defines___progname" != "xyes"])
+AM_CONDITIONAL([NEED_GETSUBOPT], [test "x$ac_cv_have_decl_suboptarg" != "xyes"])
 ##
 
 
--- a/openbsd-compat/Makefile.am
+++ b/openbsd-compat/Makefile.am
@@ -256,3 +256,7 @@ endif
 if !NEED_ERR_H
 AM_CPPFLAGS += -I$(top_srcdir)/openbsd-compat/err_h
 endif
+
+if NEED_GETSUBOPT
+libopenbsd_a_SOURCES += getsubopt.c
+endif
--- a/openbsd-compat/getsubopt.c
+++ b/openbsd-compat/getsubopt.c
@@ -1,0 +1,92 @@
+/*	$OpenBSD: getsubopt.c,v 1.4 2005/08/08 08:05:36 espie Exp $	*/
+
+/*-
+ * Copyright (c) 1990, 1993
+ *	The Regents of the University of California.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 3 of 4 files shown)


timmydo:
When compiling on Linux getsubopt was found but I hit an issue in this code:

```
if (suboptarg) fatalx("invalid TLS option \"%s\"", suboptarg);
			fatalx("missing TLS option");
```
suboptarg was not defined, but I saw getsubopt in my stdlib /gnu/store/w7fp5xk2s9djdriflyhhva21b536vdaf-glibc-2.33/include/stdlib.h:

```
#if defined __USE_XOPEN_EXTENDED || defined __USE_XOPEN2K8
/* Parse comma separated suboption from *OPTIONP and match against
   strings in TOKENS.  If found return index and set *VALUEP to
   optional value introduced by an equal sign.  If the suboption is
   not part of TOKENS return in *VALUEP beginning of unknown
   suboption.  On exit *OPTIONP is set to the beginning of the next
   token or at the terminating NUL character.  */
extern int getsubopt (char **__restrict __optionp,
		      char *const *__restrict __tokens,
		      char **__restrict __valuep)
     __THROW __nonnull ((1, 2, 3)) __wur;
#endif
```

satan1st:
Thats the reason for this patch. The OpenBSD getsubopt(3) implementation
has this suboptarg extention. I have seen this extention only on BSD
systems. On the linux distros I have tested getsubopt(3) is only defined
when _XOPEN_SOURCE or _POSIX_C_SOURCE is set. While on BSD systems it is
alwais defined.

Your comment shows that the guards are also implemented in your header.
So can you share build options and best the full build log? Also some
other informations about the system (I guess it's a NixOS).

Best whay to fix this is a better check in the configure script.

timmydo:
Here is how it's being compiled on my machine. (For some reason I need to invoke make with `CFLAGS=-ltls` for because -lssl by itself doesn't seem to grab all the symbols from libressl.)

```
gcc -DHAVE_CONFIG_H -I. -I../../../mk/smtp -I../..  -I../../../usr.sbin/smtpd -I../../../openbsd-compat -I../../../openbsd-compat/err_h  -I../../../mk/smtp  -DSMTPD_CONFDIR=\"/usr/local/etc\" -DPATH_CHROOT=\"/var/empty\" -DPATH_SMTPCTL=\"/usr/local/sbin/smtpctl\" -DPATH_MAILLOCAL=\"/usr/local/libexec/opensmtpd/mail.local\" -DPATH_MAKEMAP=\"/usr/local/sbin/makemap\" -DPATH_LIBEXEC=\"/usr/local/libexec/opensmtpd\" -DHAVE_CONFIG_H -DIO_TLS -ltls -c -o ../../usr.sbin/smtpd/smtp-smtpc.o `test -f '../../usr.sbin/smtpd/smtpc.c' || echo '../../../mk/smtp/'`../../usr.sbin/smtpd/smtpc.c
../../../mk/smtp/../../usr.sbin/smtpd/smtpc.c: In function ‘parse_tls_options’:
../../../mk/smtp/../../usr.sbin/smtpd/smtpc.c:117:23: error: ‘suboptarg’ undeclared (first use in this function); did you mean ‘optarg’?
  117 |                   if (suboptarg)
      |                       ^~~~~~~~~
      |                       optarg
../../../mk/smtp/../../usr.sbin/smtpd/smtpc.c:117:23: note: each undeclared identifier is reported only once for each function it appears in
make[3]: *** [Makefile:505: ../../usr.sbin/smtpd/smtp-smtpc.o] Error 1
```

If I go to that directory and add `-E -dM` to gcc, I see
```
#define _POSIX_C_SOURCE 200809L
```

I'm all for your patch to fix this issue, but I'm not a maintainer and just started trying out OpenSMTPD recently. I was just adding my personal experience that I had trying to compile from git HEAD yesterday.

satan1st:
I have fixed the check in configure.

Not yet tested on a BSD, will do later.

jirutka:
This is not enough, you have to also add:

```diff
diff --git a/configure.ac b/configure.ac
index 18346ec2..0a09caea 100644
--- a/configure.ac
+++ b/configure.ac
@@ -2067,6 +2067,7 @@ AM_CONDITIONAL([NEED_GETSUBOPT], [test "x$ac_cv_have_decl_suboptarg" != "xyes"])
 
 AM_COND_IF([NEED_PROGNAME], [AC_DEFINE([NEED_PROGNAME], [1], [Define to 1 if NEED_PROGNAME])])
 AM_COND_IF([NEED_SETPROCTITLE], [AC_DEFINE([NEED_SETPROCTITLE], [1], [Define to 1 if NEED_SETPROCTITLE])])
+AM_COND_IF([NEED_GETSUBOPT], [AC_DEFINE([NEED_GETSUBOPT], [1], [Define to 1 if NEED_GETSUBOPT])])
 
 
 AC_CONFIG_FILES([Makefile
```

PR #1177, opened 2022-04-26
use AC_PATH_TOOL to find ar

squalus:
This will fix detection of ar when cross compiling.

Related: https://github.com/NixOS/nixpkgs/pull/170379


orbea:
@squalus I think PR https://github.com/OpenSMTPD/OpenSMTPD/pull/1199 would be a better way of solving this?

Gentoo also has this issue report. https://bugs.gentoo.org/720782

PR #1192, opened 2023-03-31
Avoid truncation of filtered SMTP Data to 1997 bytes.

JoachimSchneider:
When using smtpd filters (e.g. rspamd) data lines get truncated after 1997 bytes: When smtpd reads the lines from the filter it truncates them because the character buffer has only LINE_MAX (2048 Bytes on OpenBSD, FreeBSD and Linux).
This patch avoids this truncation.
Greetings,
Joachim

### DIFF
--- a/usr.sbin/smtpd/lka_filter.c
+++ b/usr.sbin/smtpd/lka_filter.c
@@ -35,6 +35,8 @@
 
 #define	PROTOCOL_VERSION	"0.6"
 
+#define RESPONSE_LEN_MAX	(SMTP_LINE_MAX + 256)
+
 struct filter;
 struct filter_session;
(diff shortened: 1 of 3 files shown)


omar-polo:
Sorry for the delay and thanks for the report.

I think it's wasteful to allocate a buffer that big on the stack (SMTPD_LINE_MAX is 65535 bytes.)  instead, we could rework the function to parse the `line` in place, advancing the pointer.

I've sent a diff to tech@: https://marc.info/?l=openbsd-tech&m=168729002530904&w=2

I'll reply here too with the updates.

Thanks!

omar-polo:
After the feedback on tech@, I've committed a diff that address this issue.  As I was saying, instead of enlarging the buffer we can avoid the copying at all and just advance the string as needed.

cheers,

JoachimSchneider:
Hello Omar,

thank you for the update.

Greetings,
Joachim

PR #1195, opened 2023-04-13
build: Fix -Werror=implicit-function-declaration

orbea:
This has two commits.

1.

On a system with musl these functions are not available, but they are found by the build system inside of libbsd instead. However many of the relevant headers are never incuded resulting in many implicit function declarations. Additionally clang-16 is more strict about these turning them into errors.

2.

Split into PR https://github.com/OpenSMTPD/OpenSMTPD/pull/1198.

### DIFF
--- a/openbsd-compat/openbsd-compat.h
+++ b/openbsd-compat/openbsd-compat.h
@@ -92,6 +92,10 @@ int	scan_scaled(char *, long long *);
 const char *inet_ntop(int af, const void *src, char *dst, socklen_t size);
 #endif
 
+#ifndef HAVE_RES_RANDOMID
+unsigned int res_randomid(void);
+#endif
+
 #ifndef HAVE_STRSEP
 char *strsep(char **stringp, const char *delim);
 #endif
--- a/usr.sbin/smtpd/mail.maildir.c
+++ b/usr.sbin/smtpd/mail.maildir.c
@@ -31,6 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sysexits.h>
+#include <time.h>
 #include <unistd.h>
 
 #define	MAILADDR_ESCAPE		"!#$%&'*/?^`{|}~"
--- a/usr.sbin/smtpd/parse.y
+++ b/usr.sbin/smtpd/parse.y
@@ -44,6 +44,8 @@
 #include <netdb.h>
 #include <pwd.h>
 #include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
 #include <resolv.h>
 #include <syslog.h>
 #include <unistd.h>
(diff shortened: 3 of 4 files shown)


poolpOrg:
hello, I reviewed and while it's generally ok, you should split into two bits:

- the changes to code that are "generic" in the sense that they apply to OpenBSD, that is for example changing a function prototype from foo() to foo(void): it's ok, but should be a different diff, probably applied before this one.

- the remaining which is 100% portable related and can be merged freely here without creating a delta with upstream

orbea:
I split the two commits into different PRs.

Where should the patch against OpenBSD be sent? I do not have an OpenBSD install so I have never contributed to them other than testing some of their software.

orbea:
Thanks, I didn't know about that and will give it a try.

omar-polo:
@orbea note that got-portable makes mandatory to use libbsd when building on linux.  I'm not sure what @poolpOrg prefers to do here, but I'd start with an optional check (early, before the various AC_CHECK_HEADERS etc) and not failing if not found.

orbea:
@omar-polo and @poolpOrg I removed the libbsd specific fixes from this PR for only the more general implicit function declarations. I will make a new PR for libbsd-overlay.

And agreed, it should probably be optional.

orbea:
I made the libbsd-overlay PR here. https://github.com/OpenSMTPD/OpenSMTPD/pull/1203

omar-polo:
Not sure.  While the missing `string.h` and `time.h` are (i believe) obviously missing in OpenBSD too, setproctitle is in `stdlib.h` in OpenBSD and NetBSD.  Is FreeBSD that has it in `unistd.h` so that has to remain in -portable.  Maybe I'd tweak the comment to explain it, dunno.

orbea:
Thanks for explaining the details, I wonder why FreeBSD and NetBSD differ from OpenBSD in this regard? Would including `unistd.h` be harmful for OpenBSD here? I don't feel strongly about which repo this belongs in, but from past experience in other projects it is possible to support multiple operating systems including both linux and most of the applicable BSDs within a single code base without too much trouble. Of course I can understand if OpenBSD finds that out of scope for their repo.

poolpOrg:
I see a LGTM from @omar-polo, should this be merged ?

orbea:
@poolpOrg I think its ready, it fixes many of the build issues on my Gentoo musl system, the other build issues are in other PRs.

omar-polo:
@poolpOrg yes, please.  The discussion slightly derailed, but this pr looks fine to me and i'll try to get it committed in OpenBSD as well.  Thanks!

omar-polo:
@orbea FYI: I've just committed upstream the addition of those few missing include (and other few nits to reduce the diff with -portable.)  It should show up on the git mirror in a couple of hours.

PR #1196, opened 2023-04-13
configure.ac: Check for crypt_gensalt()

orbea:
On musl systems `crypt()` is provided by `unistd.h` and it is marked as `"none required"` during the configure process. This results in `-lcrypt` never being passed to the compiler and then an undefined reference for `crypt_gensalt()`. This can be solved by checking for `crypt_gensalt()` instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -543,11 +543,11 @@ AC_SEARCH_LIBS([setsockopt],
 			[Define if you have the setsockopt() function.])
 	])
 
-AC_SEARCH_LIBS([crypt],
+AC_SEARCH_LIBS([crypt_gensalt],
 	[crypt],
 	[
-		AC_DEFINE([HAVE_CRYPT], [1],
-			[Define if you have the crypt() function.])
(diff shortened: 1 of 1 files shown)


orbea:
Oh, I now see. musl provides a crypt function in `/usr/include/unistd.h` which doesn't require specific linking.
```
char *crypt(const char *, const char *);
```
Where `crypt_gensalt()` is not provided by musl and must be linked explicitly. I am unsure if this is the way you would prefer to handle this?

orbea:
@poolpOrg I rebased this with no changes and its still an issue for me.

omar-polo:
@orbea just fyi: i had to re-add the `crypt()` check because this broke the build on FreeBSD (who weirdly has `crypt()` in `unistd.h` but requires `-lcrypto` and doesn't have `crypt_gensalt()`.)

not a big deal, but I wanted to notify so that if you see the configure looking for `crypt` you know :-)

orbea:
@omar-polo Good to know, thanks for the ping.

PR #1197, opened 2023-04-13
configure.ac: Check for libtls

orbea:
When building OpenSMTPD with LibreSSL there will be undefined references for tls functions because LibreSSL has these as part of libtls where `-ltls` is missing from the build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -569,6 +569,13 @@ AC_SEARCH_LIBS([event_asr_run],
 			[Define if you have the event_asr_run() function.])
 	])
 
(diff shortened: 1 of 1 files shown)


poolpOrg:
just a quick note to tell you that I'll review your merge requests this week-end as I'm currently in training

orbea:
Thanks, its appreciated. After looking at issue https://github.com/OpenSMTPD/OpenSMTPD/issues/1193 I was unsure if PRs would be reviewed. :)

PR #1198, opened 2023-04-17
build: Fix -Werror=strict-prototypes

orbea:
Clang-16 is more strict about these so make the compiler happy.

Split from PR https://github.com/OpenSMTPD/OpenSMTPD/pull/1195.

### DIFF
--- a/usr.sbin/smtpd/bounce.c
+++ b/usr.sbin/smtpd/bounce.c
@@ -243,7 +243,7 @@ bounce_timeout(int fd, short ev, void *arg)
 }
 
 static void
-bounce_drain()
+bounce_drain(void)
 {
 	struct bounce_message	*msg;
 	struct timeval		 tv;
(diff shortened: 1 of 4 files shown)


orbea:
@poolpOrg These are already applied upstream.

https://github.com/openbsd/src/blob/00d833503e2654131fa16df4558102cba7eb2804/usr.sbin/smtpd/bounce.c#L239
https://github.com/openbsd/src/blob/00d833503e2654131fa16df4558102cba7eb2804/usr.sbin/smtpd/ioev.c#L227
https://github.com/openbsd/src/blob/00d833503e2654131fa16df4558102cba7eb2804/usr.sbin/smtpd/mda.c#L373
https://github.com/openbsd/src/blob/00d833503e2654131fa16df4558102cba7eb2804/usr.sbin/smtpd/mda.c#L378
https://github.com/openbsd/src/blob/00d833503e2654131fa16df4558102cba7eb2804/usr.sbin/smtpd/smtpd.c#L311

orbea:
@poolpOrg I added a commit to fix one more instance I missed before in `configure.ac`.

poolpOrg:
This is no longer necessary with the upstream sync done by @omar-polo with #1201 , care to confirm ?

PR #1201, opened 2023-05-08
sync with CVS

omar-polo:
This should bring OpenSMTPD-portable in pair with -CURRENT.  So far I've only build-tested it on OpenBSD and void linux (musl -- needed libbsd too) but as I was about to set up a new mx these days, i'll be able to test it more throughfully on linux soon.

A change from deraadt@ is still missing, but since it's tiny and allows to drop checks from the configure too I'll backport it in a separate PR once this lands: https://github.com/openbsd/src/commit/e16a722a8a965b85c7889b97d3d2e0580039ade0#diff-7e16a1fb5fcbd028a0757ac4fb5f020766f91e02fd50baf2fa5df6cc67ffb82e


poolpOrg:
nice

PR #1205, opened 2023-05-09
sync openbsd-compat with openssh-portable and base

omar-polo:
This brings in various improvements: from typos, to bugfixes to new functions.  Most notably omission for the moment is arc4random that requires more attention and will be sync'ed soon.


LeonMelis:
I have nothing to add to this PR, but I just wanted to say **thank you** for breathing new life into this repo! 

I was afraid that this repo  (mirror?) was stale indeed, and that Linux users would be left behind. This is a step in the right direction :-)

poolpOrg:
Thanks a lot @omar-polo <3

PR #1206, opened 2023-05-10
fix USER and LOGNAME in the env of the mda

omar-polo:
spotted while looking at the diff with CVS, it was changed by crisz@ in rev. 1.7 with

> Run lmtp deliveries as SMTPD_USER instead of the recipient user.

which was ported, since the other part of that commit (parse.y bits) are in-tree.  I couldn't find in the history when this was exactly backported, sorry!

@poolpOrg just to be sure, can you please take a look?

### DIFF
--- a/usr.sbin/smtpd/mda_unpriv.c
+++ b/usr.sbin/smtpd/mda_unpriv.c
@@ -64,8 +64,8 @@ mda_unpriv(struct dispatcher *dsp, struct deliver *deliver,
 	xasprintf(&mda_environ[idx++], "RECIPIENT=%s@%s", deliver->dest.user, deliver->dest.domain);
 	xasprintf(&mda_environ[idx++], "SHELL=/bin/sh");
 	xasprintf(&mda_environ[idx++], "LOCAL=%s", deliver->rcpt.user);
-	xasprintf(&mda_environ[idx++], "LOGNAME=%s", pw_name);
-	xasprintf(&mda_environ[idx++], "USER=%s", pw_name);
+	xasprintf(&mda_environ[idx++], "LOGNAME=%s", deliver->userinfo.username);
+	xasprintf(&mda_environ[idx++], "USER=%s", deliver->userinfo.username);
 
 	if (deliver->sender.user[0])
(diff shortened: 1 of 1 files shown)


poolpOrg:
This change happened after I left the project and I have no idea why :-/

Before this diff, LMTP deliveries happened as the recipient user so that delivery to me would happen as gilles, now they will happen as _smtpd, which seems like a regression unless there's a good rationale behind

omar-polo:
I think I've found the thread on tech@ about this change: https://marc.info/?l=openbsd-tech&m=158789141206143&w=2

> The LMTP change seems interesting to me, it means that a broken LMTP delivery will fail with _smtpd privileges instead of the (unprivileged) recipient user so I think it's a good move.

which admittedly doesn't seem bad, as it restricts what the LMTP delivery can do to the file system (permission-wise) to the _smtpd user instead of the recipient.

poolpOrg:
I don't know, it seems to have convinced me in 2020 when I was still heavily hacking on that code so I guess I'm missing a point nowadays ;-)

purevg:
Hello,

Since I upgraded opensmtpd from a 6.8 version to 7.3 version, I stumbled upon delivery errors through lmtp. The issue comes from the line `dsp->u.local.user = SMTPD_USER;` being added for lmtp delivery in parse.y.

The problem with this change is that in my system restricts access to lmtp socket to a specific group (and is owned by a different user than the SMTPD_USER).

But adding the SMTPD_USER to the group will not do anything as all supplementary groups are dropped while forking. ( `setgroups(1, &pw_gid)`  in smtpd.c). And adding "user" to the user owning the socket to the action line in the configuration file is refused since the dispatcher has already a user defined.

Is there something I'm missing to be able to allow the smtpd process to access my lmtp socket ?

Thank you

purevg:
I just thought of a workaround. I set mail.lmtp program owner to the SMTPD_USER and the group to the user having write access to the lmtp socket and set the mode to 2550. It allows only the opensmtpd process to use the mail.lmtp program and to have its group set to the correct group. It is working fine. This solution is enough for me.

Issue #1171, opened 2022-04-09
Incompatible with OpenSSL 3.0


kj:
Since updating to Ubuntu 22.04 prerelease (OpenSMTPD 6.8.0p2), where Ubuntu has switched from OpenSSL 1.1 to OpenSSL 3.0, I am no longer able to use OpenSMTPD with TLS connections, so I am unable to send mail. They fail with a message like:

```
1f7e23624a5a1c6c smtp connected address=123.123.123.123 host=example.com
debug: looking up pki "example.com"
debug: session_start_ssl: switching to SSL
debug: pony: ecdsae_sign
debug: pony: ecdsae_do_sign
debug: SSL library error: io_dispatch_accept_tls:SSL_accept: error:0800007D:elliptic curve routines::missing private key
debug: SSL library error: io_dispatch_accept_tls:SSL_accept: error:0A080006:SSL routines::EVP lib
1f7e23624a5a1c6c smtp disconnected reason="io-error: error:0A080006:SSL routines::EVP lib"
```

Plain text connections work fine. I don't have any experience with OpenSSL, so I'm not sure what the problem is exactly, but I guess that some API changed. There are two issues that seem potentially relevant:

https://github.com/openssl/openssl/issues/17614
https://github.com/openssl/openssl/issues/18028

Possibly more:

https://github.com/openssl/openssl/issues?q=is%3Aissue+is%3Aopen+%22missing+private+key%22

And there is a migration guide here (way over my head):

https://www.openssl.org/docs/man3.0/man7/migration_guide.html

petrovr:
Looks like doubling with #17286. Another issue #17092 is for "engines" but reason look similar.

DemiMarie:
It looks like OpenSMTPD will need to implement a custom provider.

petrovr:
> It looks like OpenSMTPD will need to implement a custom provider.

No! Just arrange code to set key method in the way that throw out misfunctional OpenSSL 3+ "key manager" (https://gitlab.com/secsh/pkixssh/-/commit/1ef9e2b6f30b05b4bfcb59607e4c31ec460fceeb).

Unfortunately I down not have enough time to finish tests with key obtained from X.509 certificate, but idea is clear.

Regards,
Roumen Petrov

sgbeal:
[...]

PR #1207, opened 2023-05-10
sync openbsd-compat/libtls with latest libtls

omar-polo:
This brings in various (small) improvements to the libtls api since the import in OpenSMTPD-portable.  pthread and the signer APIs are still left out, but otherwise the diff with the in-tree version is almost only headers.

It's still WIP because there are currently a few more things I'd need to take a closer look at (code hidden behind SUPPORT_ECDSA in particular) and because I need to properly test at runtime.


poolpOrg:
sure, and feel free to hit me up on my discord if you want to chat about SUPPORT_ECDSA, it is tricky and you'll probably need guidance to avoid a lot of pain ;-)

omar-polo:
eheh, thanks!  Unfortunately I don't have discord*, but I can reassure you I'm starting to get an idea of what amount of pain you're referring to :D

*: hope email / irc / icb is fine too in case?

just fyi: i've committed upstream one of the local edits you had in tls.c: https://github.com/openbsd/src/commit/3130d2a2d699478c599513e90c53b911843ac7df

poolpOrg:
E-mail will be a bit harder than a chat, if you're on IRC let me know where to join and times you're around and I'll join

omar-polo:
hmpf... I had to lower `TLS_CIPHERS_DEFAULT` to `TLS_CIPHERS_COMPAT` otherwise `tls_config_new` always fails in `tls_config_set_ciphers`.

libretls (the libtls implementation targetting OpenSSL) also has this, I've actually stolen that from them, but meh :/

I've tested locally.  With this I'm able to send email from OpenBSD (host) to a vmd guest (void linux, musl with OpenSSL 1.1) using a self-signed cert.  A similar setup but with an alpine vmd guest (musl, OpenSSL 3.1) doesn't work at runtime, but that's nothing new, it's the same issue as #1171 that i still have to look into.

if you're fine, I'd merge this and start looking into the issues with OpenSSL 3.x tomorrow.

poolpOrg:
sure

omar-polo:
Thanks! :)

PR #1208, opened 2023-05-11
libtls: unbreak with OpenSSL 3.x

omar-polo:
In OpenSSL 3.x due to a behaviour change EVP_PKEY_get1_EC_KEY() returns a cashed copy of the provider's key, so subsequent updates are not reflected back breaking the privsep usage and making the handshake fail with a cryptic "missing private key".  Instead, we have to set the changed key explicitly.

See #1171


omar-polo:
p.s. this will need a little tweak once i'll add the signer stuff to the bundled libtls to shorten the diffs with libtls...

poolpOrg:
fine by me

omar-polo:
Thanks! 

I'm also discussing a similar patch for libtls, once it gets committed i'll forward the patch to the libretls maintainers too.  I'm also about to add a configure knob `--with-bundled-libtls` to avoid using the system version, something that will be needed until libretls gets updated.

PR #1210, opened 2023-05-25
add missing include of sys/time.h and time.h

omar-polo:
@dfateyev could you please give this PR a spin on RHEL7?  Should fix the build failure, although I won't be surprised if it wants `time.h` in smtpd.h.

Thanks!

### DIFF
--- a/usr.sbin/smtpd/bounce.c
+++ b/usr.sbin/smtpd/bounce.c
@@ -29,6 +29,7 @@
 #include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "smtpd.h"
--- a/usr.sbin/smtpd/control.c
+++ b/usr.sbin/smtpd/control.c
@@ -35,6 +35,7 @@
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "smtpd.h"
--- a/usr.sbin/smtpd/enqueue.c
+++ b/usr.sbin/smtpd/enqueue.c
@@ -32,6 +32,7 @@
 #include <pwd.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "smtpd.h"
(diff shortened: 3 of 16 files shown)


omar-polo:
oops, forgot to commit a tweak to the master branch, had to force-push.

dfateyev:
New (failed) build log: https://kojipkgs.fedoraproject.org//work/tasks/2088/101532088/build.log

omar-polo:
Can't say I'm not tempted, but the truth is that gcc was right here.  We were lacking `sys/time.h` and still lacking `time.h` in various other places; we're just lucky that on the system we've tested so far those headers gets pulled in by other headers, so I think there's some value in this.

(Said that, I'm still surprised it errors in to.c, localtime(3) is provided by time.h which this patch adds...)

What is driving me mad is my inability of understanding how to try to replicate myself.  I've created a container based on rhel7, but I'm unable to install stuff in it.  From what I've understood I need to run `subscription-manager` to enable other repositories (like `rhel-server-devtools-7-rpms`) but it's disabled when running inside a container.  Can you please help me with this?  I've used fedora in the past, but RHEL is still out of my comfort zone.

dfateyev:
> What is driving me mad is my inability of understanding how to try to replicate myself. I've created a container based on rhel7, but I'm unable to install stuff in it. From what I've understood I need to run `subscription-manager` to enable other repositories (like `rhel-server-devtools-7-rpms`) but it's disabled when running inside a container. Can you please help me with this? I've used fedora in the past, but RHEL is still out of my comfort zone.

You can install [Centos7](https://www.centos.org/centos-linux/) — it's free and 100% binary compatible with RHEL7.
You don't need `rhel-server-devtools-7-rpms`, just install OS and add EPEL repository: `yum install epel-release` and "rpm-build": `yum install rpm-build` to be able to (re-)build packages.

After that, you can:
1) Grab and install testing source RPM: `rpm -ivh https://kojipkgs.fedoraproject.org//work/tasks/6715/101536715/opensmtpd-7.0.0p1-1.20230525.el8.src.rpm`
2) Run package build from the installed source RPM above: `rpmbuild -ba ~/rpmbuild/SPECS/opensmtpd.spec`. It will tell which devel packages are needed additionally. After installing them, the build process will start, and you can check the errors during the build.
3) To update sources, just update `~/rpmbuild/SOURCES/opensmtpd-master.zip` with newer version.

omar-polo:
Thanks for the guidance!  I've updated the PR and merged, I'll take care of upstreaming it too.

PR #1211, opened 2023-06-01
start to sync libasr with base

omar-polo:
lots of changes imported in one go.  This notably excludes the commit from florian@ to consider the current routing domain.

### DIFF
--- a/openbsd-compat/libasr/asr.c
+++ b/openbsd-compat/libasr/asr.c
@@ -165,12 +165,14 @@ asr_resolver_from_string(const char *str)
 
 	return ac;
 }
+DEF_WEAK(asr_resolver_from_string);
 
 void
 asr_resolver_free(void *arg)
 {
 	_asr_ctx_unref(arg);
 }
+DEF_WEAK(asr_resolver_free);
 
 /*
  * Cancel an async query.
--- a/openbsd-compat/libasr/asr.c
+++ b/openbsd-compat/libasr/asr.c
@@ -1,4 +1,4 @@
-/*	$OpenBSD: asr.c,v 1.61 2018/10/22 17:31:24 krw Exp $	*/
+/*	$OpenBSD: asr.c,v 1.68 2022/01/20 14:18:10 naddy Exp $	*/
 /*
  * Copyright (c) 2010-2012 Eric Faurot <eric@openbsd.org>
  *
@@ -116,11 +116,15 @@ _asr_resolver(void)
 void
 _asr_resolver_done(void *arg)
 {
-	struct asr *asr = arg;
+	struct asr_ctx *ac = arg;
+	struct asr *asr;
 	struct asr **priv;
 
-	if (asr == NULL) {
-		priv = _THREAD_PRIVATE(_asr, _asr, &_asr);
+	if (ac) {
+		_asr_ctx_unref(ac);
+		return;
+	} else {
+		priv = _THREAD_PRIVATE_DT(_asr, _asr, NULL, &_asr);
 		if (*priv == NULL)
 			return;
 		asr = *priv;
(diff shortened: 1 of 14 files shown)


omar-polo:
This includes OpenBSD 7.2 errata 22 "Out of bounds accesses in libc resolver."

dfateyev:
@omar-polo thanks for the fixes! have you considered to update the separate [libasr repo](https://github.com/OpenSMTPD/libasr), as well?

omar-polo:
@dfateyev I believe @poolpOrg idea was to always build OpenSMTPD using the bundled libasr, and the libasr repo being sort-of abandoned since taking care of maintaining both that and the bundled version here was too much.  The configure check is left so that OpenSMTPD still builds on OpenBSD.

I have a couple of more pressing things to do at the moment, and won't promise anything, but I'd like to give it a try at updating the libasr repository and making a release there too in the not-so-distant future as well.

omar-polo:
@dfateyev I've talked with @poolpOrg and I agree with the direction taken: libasr will be maintained only as a part of OpenSMTPD, the libasr-portable repository will be archived soon.

I'll try to contact the maintainers of the libasr package to have it removed from their repositories, as it has several well-known issues fixed upstream, and to have OpenSMTPD build with the bundled version.

dfateyev:
> @dfateyev I've talked with @poolpOrg and I agree with the direction taken: libasr will be maintained only as a part of OpenSMTPD, the libasr-portable repository will be archived soon.

@omar-polo as for retiring libasr-portable, it sounds a kind of disappointing. I mean, the idea to make it a separate library so other projects could use its functionality is good, and fully correlates with an intention to unbundle libs from the core functionality, which we are used to in various systems.

For sure, we can deprecate current `libasr` packages in distributions, and proceed with the bundled. I understand, that a maintenance of an unversioned internal library is slightly easier, but: is it much easier? In case of changes in `libasr` code, aren't they practically the same, regardless of the library's implementation (internal or external)? Also, having `libasr` internal, it's needed to release a full `opensmtpd` package, instead of updating one library separately, as it is now.

omar-polo:
@dfateyev I understand that having libasr separate seems at first a good idea, and I'm in general against bundling libraries for the exact reasons you mention, but libasr is a particular case.  It's not a question of maintenance burden really, it's a case of feasibility and burden on the _users_ of the library.

libasr on OpenBSD is part of libc.  as such, it can do a few interesting things that a separate library can't do like building objects to be free'd via standard functions.  For example, when using `getaddrinfo_async()`, `asr_run()` will create a series of `struct addrinfo` that can later be passed to libc' `freeaddrinfo()`, exactly like the ones you get from the standard `getaddrinfo()`.  In a portable library we can't do this, and in fact we had to introduce `asr_freeaddrinfo()` and change the code accordingly.

Another important issue is the libevent integration.  The libevent 1.x you'll find in base on OpenBSD has an internal asr integration via `event_asr_run()`, something that is also difficult to do in a separate and portable "libasr" because you'll likely don't want to tie libasr to a relatively big library like libevent just for a integration bit.

Then, if you take into consideration that the user of libasr are programs targeting OpenBSD, we felt that the idea of maintaining libasr bundled and having other -portable projects copy our `openbsd-compat/libasr/` was the simplest choice.

This doesn't preclude that in the future libasr-portable could be revamped and made a separate library again.  To be fair, I'd like to see that happening, but it would take someone to untangle the points I've mentioned before in a way that doesn't make it difficult to switch back and forth between OpenBSD' asr_run and libasr-portable with or without libevent integration, and in a way that's handy for everyone.

dfateyev:
Can `libasr` static linking solve the problems that you have mentioned? If yes, it could be an integral part of opensmtpd binary(-ies), at the same time being a library that could be developed and distributed separately.

PR #1216, opened 2023-06-26
Add GH Runner to build images of release tags

SohamG:
Hello OpenSMTPD team,

I noticed that there were no "official" "source of truth" docker/container images for OpenSMTPD. I would like to propose and help have "official" opensmtpd containers on Github's container registry `ghcr.io/opensmtd/<image>`. 

Currently, I have adapted the `ci` dockerfile for alpine to add the entrypoint, and added the GH runner yaml to parse the tag number and apply it as a tag on the image (taken from github's documentation).

I assume extensive testing will be required (which I haven't gotten around to), opening this PR to open discussion.

### DIFF
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -3,6 +3,8 @@ on:
   push:
     tags:
       - v*
+env:
+  IMAGE_NAME: ${{ github.repository }}
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -2,7 +2,6 @@ name: Alpine Linux release
 on: 
   push:
     tags:
-      - *
 env:
   IMAGE_NAME: "opensmtpd"
 jobs:
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -4,14 +4,14 @@ on:
     tags:
       - v*
 env:
-  IMAGE_NAME: ${{ github.repository }}
+  IMAGE_NAME: "opensmtpd"
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
     - name: Alpine Linux (amd64 musl openssl)
-      run: docker build . --file ci/docker/Dockerfile.alpine --tag opensmtpd:alpine
+      run: docker build . --file ci/docker/Dockerfile.alpine-release --tag opensmtpd:alpine
 
     - name: Login to ghcr.io
       run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -11,11 +11,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: Alpine Linux (amd64 musl openssl)
+
+    - name: Build the Dockerfile
       run: docker build . --file ci/docker/Dockerfile.alpine-release --tag opensmtpd:alpine
 
     - name: Login to ghcr.io
-      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+      uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    #   run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+    
 
     - name: Push Image
       run: |
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -7,6 +7,7 @@ env:
   IMAGE_NAME: "opensmtpd"
 jobs:
   build:
+    permissions: write-all
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -1,0 +1,32 @@
+name: Alpine Linux (amd64 musl openssl)
+on: 
+  push:
+    tags:
+      - v*
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Alpine Linux (amd64 musl openssl)
+      run: docker build . --file ci/docker/Dockerfile.alpine --tag opensmtpd:alpine
+
+    - name: Login to ghcr.io
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+
+    - name: Push Image
+      run: |
+        IMAGE_ID=ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME
+        # Change all uppercase to lowercase
+        IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
+        # Strip git ref prefix from version
+        VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+        # Strip "v" prefix from tag name
+        [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+        # Use Docker `latest` tag convention
+        [ "$VERSION" == "master" ] && VERSION=latest
+        echo IMAGE_ID=$IMAGE_ID
+        echo VERSION=$VERSION
+        docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
+        docker push $IMAGE_ID:$VERSION
+
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -26,9 +26,8 @@ jobs:
         # Strip "v" prefix from tag name
         [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
         # Use Docker `latest` tag convention
-        [ "$VERSION" == "master" ] && VERSION=latest
         echo IMAGE_ID=$IMAGE_ID
         echo VERSION=$VERSION
-        docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
+        docker tag $IMAGE_NAME:alpine $IMAGE_ID:$VERSION
         docker push $IMAGE_ID:$VERSION
 
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -1,4 +1,4 @@
-name: Alpine Linux (amd64 musl openssl)
+name: Alpine Linux release
 on: 
   push:
     tags:
--- a/.github/workflows/alpine-release.yml
+++ b/.github/workflows/alpine-release.yml
@@ -2,7 +2,7 @@ name: Alpine Linux release
 on: 
   push:
     tags:
-      - v*
+      - *
 env:
   IMAGE_NAME: "opensmtpd"
 jobs:
@@ -32,10 +32,9 @@ jobs:
         # Strip git ref prefix from version
         VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
         # Strip "v" prefix from tag name
-        [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
-        # Use Docker `latest` tag convention
+        # [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
         echo IMAGE_ID=$IMAGE_ID
         echo VERSION=$VERSION
-        docker tag $IMAGE_NAME:alpine $IMAGE_ID:$VERSION
+        docker tag opensmtpd:alpine $IMAGE_ID:$VERSION
         docker push $IMAGE_ID:$VERSION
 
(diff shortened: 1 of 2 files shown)


omar-polo:
Hello,

I'm not sure if we want to provide an "official" docker image, after all we don't provide any "official" binary as well (not trying to say we should, I believe it's not the developer' job to provide binaries). I'm neuter to either providing something like this or an "official" Dockerfile for actual reuse, I'll leave the decision to @poolpOrg.

For what is worth, I can help do some testing.  I'm already running one OpenSMTPD instance on linux under docker (podman really) using a similar Dockerfile.

It's very likely that one need to run multiple things inside the OpenSMTPD container (the filters needs to stay in the same container AFAIK), but it's easier to just use `from OpenSMTPD` and add stuff on top eventually.

I'll need some time to properly review this as I don't have experience with using the github registry and automatizing it, so far I've only used private self-hosted registries with a self-hosted CI.  I'll leave some comments from an initial read however.


Thanks,

.github/workflows/alpine-release.yml
```diff
@@ -0,0 +1,34 @@
+name: Alpine Linux release
+on: 
+  push:
+    tags:
+      - v*
```
omar-polo, line 5: If I'm reading this correctly (i.e. using tags that match `v*`) then it's wrong.  We haven't been consistent so far with the tag naming, but I'll stick to just using the version number from now on (i.e. 7.3.0p0, 7.4.0p0, etc...) without any prefix.

.github/workflows/alpine-release.yml
```diff
@@ -0,0 +1,34 @@
...
+  build:
+    permissions: write-all
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Alpine Linux (amd64 musl openssl)
+      run: docker build . --file ci/docker/Dockerfile.alpine-release --tag opensmtpd:alpine
+
+    - name: Login to ghcr.io
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
```
omar-polo, line 18: the `-u $` seems strange, forgot to type the variable' name?

.github/workflows/alpine-release.yml
```diff
@@ -0,0 +1,34 @@
...
+        IMAGE_ID=ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME
+        # Change all uppercase to lowercase
+        IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
+        # Strip git ref prefix from version
+        VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+        # Strip "v" prefix from tag name
+        [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+        # Use Docker `latest` tag convention
+        echo IMAGE_ID=$IMAGE_ID
+        echo VERSION=$VERSION
```
omar-polo, line 31: These two are just for debugging / providing a nicer output or forgot to `export` / continue the pipeline?

ci/docker/Dockerfile.alpine-release
```diff
@@ -0,0 +1,51 @@
...
+    autoconf \
+    automake \
+    bison \
+    ca-certificates \
+    fts-dev \
+    gcc \
+    fts \
+    libevent-dev \
+    libtool \
+    libtool \
```
omar-polo, line 18: This is an issue with the current Dockerfile as well, but there's no need to specify `libtool` twice.  Actually, the exact list of dependencies should be re-evaluated probably, but that's a separate issue from this PR.

ci/docker/Dockerfile.alpine-release
```diff
@@ -0,0 +1,51 @@
...
+    linux-pam-dev \
+    make \
+    musl-dev \
+    openssl \
+    openssl-dev \
```
omar-polo, line 23: not sure how easy is to run with libressl on alpine but I'd say that if we start providing an "official" dockerfile it should be linked to libressl, by default at least.  The CI is fine using OpenSSL as it's for testing.

This also brings the question of how to automatically push a new image when OpenSSL / LibreSSL gets updated.  The other dependencies are not so important, but updates on OpenSSL and LibreSSL (and maybe libevent) should be applied as soon as possible.

ci/docker/Dockerfile.alpine-release
```diff
@@ -0,0 +1,51 @@
...
+    zlib-dev
```
omar-polo, line 24: Last thing, if we start providing an "official" Dockerfile I'd prefer if it didn't contain the packages needed to build.  (have to admit that the Dockerfile I'm currently using keeps the build-dependency because I'm lazy ;-)

How hard would it be to change this into a multi-stage build?  We'd need to copy `smtpd`, `smtpctl`, `smtp`, the libexecs inside `/usr/local/libexec/opensmtpd` and re-create some directories under `/var` in the clean image, then reinstall libevent, openssl/libressl.

ci/docker/Dockerfile.alpine-release
```diff
@@ -0,0 +1,51 @@
...
+RUN mkdir -p /var/lib/opensmtpd/empty \
+  && adduser _smtpd -h /var/lib/opensmtpd/empty/ -D -H -s /bin/false \
+  && adduser _smtpq -h /var/lib/opensmtpd/empty/ -D -H -s /bin/false \
+  && mkdir -p /var/spool/smtpd \
+  && mkdir -p /var/mail \
+  && mkdir -p /etc/mail \
+  && chmod 711 /var/spool/smtpd
+
+# Copy contentes of the repo inside the container
+# https://docs.docker.com/engine/reference/builder/#copy
```
omar-polo, line 38: I'd avoid keeping this kind of commits.  It's useless to link the documentation IMHO, if one doesn't know how a Dockerfile works they can find the doc by themselves, and we can avoid littering the file with links.

it's probably an issue with the CI dockerfiles.  I'll take care to clean them up sometime.

ci/docker/Dockerfile.alpine-release
omar-polo: Ah, `libbsd` should be added to mix as well, and hopefully the image be re-pushed again when also libbsd gets updated.

omar-polo (commented):
Here's a first round of review.  Don't worry if you can't fix everything, it's already a good start!  Some of the points are not really a new problem but something that we ought to cleanup in the CI dockerfiles as well.

I'm particularly worried however about shipping outdated crypto.  I know it's not easy, but I'd like to know if there is a way to (automatically or not, I can take the burden if we decide to ship the "official" image) re-push the image when OpenSSL/LibreSSL (and maybe libevent and libbsd too at least) receive an update.

I also know it's not something usually done, but I think we can try to do better :-)

.github/workflows/alpine-release.yml
SohamG: I actually lifted that from Github's documentation. I will change to use the docker login action https://github.com/docker/login-action instead.

ci/docker/Dockerfile.alpine-release
SohamG: Oh, I wasn't focused on the build part. I'll make this fix in both dockerfiles if thats okay

ci/docker/Dockerfile.alpine-release
SohamG: I can investigate `libressl`, admittedly I am a Linux (or should I say "GNU/Linux") user and have never considered alternatives to openssl.

SohamG: As far as auto-triggering goes, we would need some signal from somewhere that a new `libressl` is available in alpine. GH actions does support triggering actions through an API/from other actions.

ci/docker/Dockerfile.alpine-release
SohamG: Yeah this was taken from the CI file. I need to do cleanup anyway, I was thinking of splitting the && shell pipelines into separate RUN instructions to be more in line with the commits/layers philosophy. I'll wanted an initial go ahead and now I'll start on that.

.github/workflows/alpine-release.yml
SohamG: Sure, that works. I saw the tag drop down in the Github UI and saw a couple vXXX tags, so I assumed thats what it was.

ci/docker/Dockerfile.alpine-release
omar-polo: Not a big deal.  It's a separate issue, so it's fine with me if you prefer to fix afterwards, and regarding the CI dockerfiles they need some small tweaks anyway.  Oh, patches and are always welcome though :)

ci/docker/Dockerfile.alpine-release
omar-polo: I phrased my initial comment poorly, sorry.  I wasn't asking you to actually do this, rather hear your opinion in how much difficult / fragile it would be.

You guessed right, with a [multi-stage build](https://docs.docker.com/build/building/multi-stage/) I meant to use a first temporary image to build OpenSMTPD (along with all the build-depends needed) and then copy the binaries into a separate image.  Recent-ish docker version make this somewhat easy, although for an application like OpenSMTPD it could be tricky as it's not a single binary.

It's just a "what if".

ci/docker/Dockerfile.alpine-release
omar-polo: Actually, there's a small issue with libressl at the moment (OpenSMTPD does a few tricks not so many applications do), so let's keep openssl.  We can revisit it later.  Please discard this point.

ci/docker/Dockerfile.alpine-release
SohamG: Thanks for introducing me to this, I will try to implement this shortly.

SohamG:
As for the discussion on providing "official" packages, my 2 cents are as follows: Container images dont come with the same implications as distribution packages. When a package maintainer packages up an application for their distro there is some (IMHO) unwritten implication of quality control/testing that the package will cohesively work with the distribution.

Container images don't have this issue, they are basically treated as standalone executables, gotten and executed "as is". The motivation for this PR was I wanted to investigate running opensmtpd in my kubernetes cluster, and I didn't want to pull some random persons image who may or may not be keeping it up to date (or ofcourse may or may not be arbitrarily modifying it).

.github/workflows/alpine-release.yml
SohamG: Fixed.

.github/workflows/alpine-release.yml
SohamG: Fixed.

ci/docker/Dockerfile.alpine-release
SohamG: Fixed in the dockerfile im adding.

ci/docker/Dockerfile.alpine-release
SohamG: Implemented multi-stage build.

ci/docker/Dockerfile.alpine-release
SohamG: Removed unnecessary comments in the dockerfile I am adding.

ci/docker/Dockerfile.alpine-release
```diff
@@ -15,37 +12,54 @@ RUN apk add --no-cache \
...
+  && mkdir -p /var/mail \
+  && mkdir -p /etc/mail \
+  && chmod 711 /var/spool/smtpd
+
+RUN apk add --no-cache libevent \
+    openssl \
+    linux-pam \
+    fts
+
+COPY --from=build /usr/local /usr/local/
```
omar-polo, line 58: this is too broad, i'd go with something like (not tested yet)

```
COPY --from=build /usr/local/libexec/opensmtpd /usr/local/libexec
COPY --from=build /usr/local/sbin/{smtpctl,smtpd} /usr/local/sbin
COPY --from=build /usr/local/bin/smtp /usr/local/bin
```

omar-polo:
Thanks a lot.  With my last comment addressed this is fine for me. (still need to hear from Gilles however, I'd prefer to have his "ok" to proceed)

I have a couple of more question still however:

 - I suppose I (or @poolpOrg) need to setup something with our accounts, can you point us to the relevant documentation?
 - there is a way to test the image it produces without it being pushed? (or pushed as private)  Apologize if I'm making a dumb question, but I we're about to release a new minor soon and I'd like to test this so it doesn't break spectacularly :)

(i guess i can just apply this on my copy of the repo under my account and see how it goes.)

I'll test your dockerfile in the following days.

Thanks!

SohamG:
These are the docs for the Github Packages feature: https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages
And the container registry: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

I wouldn't mind hopping on the IRC for fine grained concerns.

Secondly, I believe the package will be pushed as private by default (as in on your client, you would have to `docker login ghcr.io`). Repo managers can then choose to make the package (container in this case) public, or ofcourse deploy any IaM policy github supports. If you want to see it in action, you can also refer to my fork https://github.com/SohamG/OpenSMTPD/pkgs/container/opensmtpd (I'll add you as a collaborator so you can dig into it a bit)

ci/docker/Dockerfile.alpine-release
SohamG: On it. I dont speak autoconf very well so I couldnt figure out the complete list of the build results

SohamG: Should be resolved.

SohamG:
Please let me know if you want me to squash.

SohamG:
Hey, is there an update on this?

omar-polo:
sorry for the delay, busy week.

it looks good to me, i've played briefly with it and -beside some confusion due to the github UI- it seems to work as intended.

I'll do one more round of testing and merge the pr tomorrow or the day after.  No need to squash, will rebase and squash before merging.

Thanks!

SohamG:
No worries, thanks for the reply. I have added you as a collaborator to my fork which has the CI running. Feel free to poke around.

SohamG:
Thanks, looking forward to the next release when this will trigger. Feel free to reach out if you have trouble making the containers public visible

omar-polo:
Took a while to notice that the visibility was hidden ^^".  Now the package is public! \o/

It worked fine with the rc1 tag, thank you!

SohamG:
My pleasure!

PR #1222, opened 2023-07-12
rework the compat layer

omar-polo:
opening the pr just for visibility since this is taking quite some time; may not compile yet.

I'm currently working through the configure script to revamp it, drop useless checks, improve the one we have etc.  One of the plans is to drop the `libopenbsd-compat` and greatly simplify `openbsd-compat/Makefile.am` by using `AC_REPLACE_FUNCS` and libobj.


omar-polo:
Finally it builds again :-)

One of the advantages is that we don't have to build libopenbsd-compat, the configure.ac and the various Makefile.am are simpler, and the various executables are smaller since they don't all have a copy of libtls and libasr inside.  A next step would be to don't link every binary to every library (e.g. `encrypt` doesn't need to link to libevent or libssl or...)

I'm now tempted to move everything under the main Makefile.am to speed up parallel builds and

PopeNobody:
I'm wondering if there is a way to build a framework that will use the BSD Makefiles, which are both tiny and beautiful, and just interpret them appropriately for whatever platform we are on.

omar-polo:
@PopeNobody there's bmake that ships with NetBSD' make infrastructure which *should* be compatible with the one on OpenBSD.  However, we still need the checks for portability (add missing functions on the host etc), so a configure script would still be needed.  There's kristaps' oconfigure that's nice, but autotools can also be nice if used good :)

omar-polo:
I've applied the trivial fixes in the master to slim down this PR (and have them in 7.4)  What remains here is the configure rewrite which unfortunately still needs work...

omar-polo:
I've also cherry-picked the guts of this into the master branch, this now lives as experiment for post 7.4

Issue #1190, opened 2023-02-19
smtpd loop with default config when dst mx points to localhost


heinrichrebehn:
Hello all,

I stumbled upon a weird situation when trying to send mail to test@webmail.de, smtpd entered into an endless loop:

———————————————————————————————————————————
obsd-test# mail -s test test@webmail.de
test
.
EOT
obsd-test# Feb 17 16:00:01 obsd-test smtpd[74143]: f63245c93259f11c smtp connected address=local host=[obsd-test.rebehn.net](http://obsd-test.rebehn.net/)
Feb 17 16:00:01 obsd-test smtpd[74143]: f63245c93259f11c smtp message msgid=7157a411 size=363 nrcpt=1 proto=ESMTP
Feb 17 16:00:01 obsd-test smtpd[74143]: f63245c93259f11c smtp envelope evpid=7157a411f91aaef7 from=<[root@obsd-test.rebehn.net](mailto:root@obsd-test.rebehn.net)> to=<[test@webmail.de](mailto:test@webmail.de)>
Feb 17 16:00:01 obsd-test smtpd[74143]: f63245c93259f11c smtp disconnected reason=quit
2023-02-17T15:00:01.986Z obsd-test newsyslog[87010]: logfile turned over
tail: /var/log/maillog has been replaced, reopening.
2023-02-17T15:00:01.986Z obsd-test newsyslog[87010]: logfile turned over
Feb 17 16:00:02 obsd-test smtpd[74143]: f63245cd0d08e264 mta connecting address=smtp://127.0.0.1:25 host=localhost
Feb 17 16:00:02 obsd-test smtpd[74143]: f63245cd0d08e264 mta connected
Feb 17 16:00:02 obsd-test smtpd[74143]: f63245ce7265d1ed smtp connected address=127.0.0.1 host=localhost
Feb 17 16:00:02 obsd-test smtpd[74143]: f63245ce7265d1ed smtp message msgid=d2ff5ca2 size=546 nrcpt=1 proto=ESMTP
Feb 17 16:00:02 obsd-test smtpd[74143]: f63245ce7265d1ed smtp envelope evpid=d2ff5ca2aa40c428 from=<[root@obsd-test.rebehn.net](mailto:root@obsd-test.rebehn.net)> to=<[test@webmail.de](mailto:test@webmail.de)>
[...]

PR #1223, opened 2023-08-01
Nullmx

satan1st:
This implement null mx (rfc 7505). Also filter `localhost` mx entries and handles only `localhost` mx entries as null mx (see #1190).

Sorry I have currently not the time to test this, maybe someone else can test it.

### DIFF
--- a/usr.sbin/smtpd/dns.c
+++ b/usr.sbin/smtpd/dns.c
@@ -249,6 +249,7 @@ dns_dispatch_mx(struct asr_result *ar, void *arg)
 	struct dns_rr		 rr;
 	char			 buf[512];
 	size_t			 found;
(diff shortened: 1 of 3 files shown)


omar-polo:
Hi @satan1st, thanks for the PR!  Unfortunately I'm a bit busy at the moment so it'll take me a while to find the time to review, test, and forward this to tech@, I'm afraid.  I just wanted to let you know that I've seen the PR and will try to allocate time for it as soon as possible.

omar-polo:
@satan1st thanks for working on this!  It has been committed on CVS (the first part about nullmx, not the localhost bit, but that can be now addressed on tech@ easily), once I'll fix an issue on macos and release a p1 I'll backport here too.

PR #1226, opened 2023-09-13
make documentation more explicit about the auth label

Ekleog:
Hey! I just spent a bit more than 4 hours (based on my DNS delayed) trying to understand why I couldn't relay email through an external authed server, before figuring out that I actually just missed adding the auth label.

Here is a suggested change to docs which should hopefully make it more explicit, as I had been reading docs first thing and I was wondering what that "label1" was about. I probably should have checked the documentation of `host` again, but as my setup without-auth did work fine I didn't think of it.

Also, it might make sense to error out when parsing a relay line has an `auth` table defined but no label, maybe, as it'd be ill-formed? Though TBH there may already have been a warning that I missed, as I thought it was a problem of networking between my node and the other node and spent most of my time debugging in that direction.

### DIFF
--- a/usr.sbin/smtpd/table.5
+++ b/usr.sbin/smtpd/table.5
@@ -153,7 +153,11 @@ label1	user:password
 .Ed
 .Pp
 The label must be unique and is used as a selector for the proper credentials
-when multiple credentials are valid for a single destination.
+when multiple credentials are valid for a single destination. This means, that
+the relay must with this example be specified like:
+.Bd -unfilled -offset indent
+.Ic action Ar name Cm relay host Ar smtp+tls://label1@example.org Cm auth Pf < Ar table Ns >
+.Ed
 The password is not encrypted as it must be provided to the remote host.
 .Ss Netaddr tables
 Netaddr tables are lists of IPv4 and IPv6 network addresses.


Ekleog:
Oh also, disclaimer: I have literally never learned the man page syntax, just copy-pasted what looked like a reasonably similar snippet and changed it in hopes that my guesses actually result in the proper manpage syntax :)

omar-polo:
Hi,

I need to take a look at why we currently accept a `relay host auth` without a label in the address.

Regarding the the manpage change I think too we should provide more clues of why the label is important and how it's linked to the address.

However, instead of adding a second config example, what about improving the the one we already have in that section?

```diff
--- usr.sbin/smtpd/table.5
+++ usr.sbin/smtpd/table.5
@@ -132,7 +132,7 @@ Credentials tables are mappings of credentials.
 They can be used in two contexts:
 .Bd -unfilled -offset indent
 .Ic listen on Ar interface Cm tls Oo Ar ... Oc Cm auth Pf < Ar table Ns >
-.Ic action Ar name Cm relay host Ar relay-url Cm auth Pf < Ar table Ns >
+.Ic action Ar name Cm relay host Ar proto Ns :// Ns Ar label Ns @ Ns Ar host Ns Oo : Ns Ar port Oc Cm auth Pf < Ar table Ns >
 .Ed
 .Pp
 In a listener context, the credentials are a mapping of username and encrypted
```

This renders as

```
   Credentials tables
     Credentials tables are mappings of credentials.  They can be used in two
     contexts:

           listen on interface tls [...] auth <table>
           action name relay host proto://label@host[:port] auth <table>
```

The markup is not great (it's an overlong line I don't know how to split), I'll have to ask for help on tech@ about that, but what do you think about the idea?


P.S.: for most common mandoc errors there's `man -Tlint` can help, I've found it really useful to adjust manpages.

omar-polo:
> I need to take a look at why we currently accept a relay host auth without a label in the address.

If I understand correctly, it's because `relay host example.com` implicitly creates a dynamic table holding `example.com`, so when we reach the `auth` we don't have a single URL but a table.

Tables are dynamic and may be backed by an external process (like a database), so checking the syntax of every entry is impossible.

We could however special-case the dynamic tables (i.e. the ones written inline in the config file or loaded file `file:`) and check for the label presence there.

Ekleog:
> However, instead of adding a second config example, what about improving the the one we already have in that section?

The reason was that I wanted it to be as close to the table example as possible, so that it's easy to see that `label1` must be the same in both the auth table and the relay label. Maybe it'd be even better to have an example with two labels in the auth table, to make this very clear?

Anyway, the fact that the two labels need to match was actually what

Ekleog:
I have completely forgotten about the context of this PR, so let's just close it, please anyone feel free to reuse the work here and complete the documentation :)

PR #1228, opened 2023-09-28
sync libtls, september 2023 edition

omar-polo:
with this we're in sync with the latest libtls from CVS, minor the pthread bits we don't use.  (including the signer APIs this time!)

### DIFF
--- a/openbsd-compat/Makefile.am
+++ b/openbsd-compat/Makefile.am
@@ -32,6 +32,7 @@ libopenbsd_a_SOURCES += libtls/tls_keypair.c
 libopenbsd_a_SOURCES += libtls/tls_server.c
 libopenbsd_a_SOURCES += libtls/tls_ocsp.c
 libopenbsd_a_SOURCES += libtls/tls_peer.c
+libopenbsd_a_SOURCES += libtls/tls_signer.c
 libopenbsd_a_SOURCES += libtls/tls_util.c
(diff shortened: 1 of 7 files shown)


whataboutpereira:
> [Remove the ability to do tls 1.0 and 1.1 from libtls.]

Does this mean TLSv1 and v1.1 will be defunct in OpenSMTPD? I'm still seeing the occasional TLSv1 or v1.1 SMTP connection.

omar-polo:
@whataboutpereira
> Does this mean TLSv1 and v1.1 will be defunct in OpenSMTPD?

Yep.

TLSv1.0 and 1.1 are a "MUST NOT use" for more than two years already (see [RFC 8996](https://www.rfc-editor.org/rfc/rfc8996)).  This is already the case on OpenBSD-current (beck@ removed the ability to do tls 1.0 and 1.1 from libtls in early August) and so also on OpenBSD 7.4+.

Also, for the archive, note that even if you have `protocols "tlsv1.1"` in your config it'll be silently ignored.

However, this shouldn't mean a loss of email from the sender that are still using ancient TLS versions.  Once the TLS handshake fails they *should* re-try in plaintext.  Not great but we're talking of ancient software that can't speak tls 1.2 nor 1.3...  Most mail servers should at least support tls 1.2 "thanks" to the power of the oligopoly.

Issue #1232, opened 2023-10-27
build errors on macOS 14.2


artkiver:
Hi there,

I am not the MacPorts OpenSMTPD maintainer, yet I thought I would give a try at updating the MacPorts' Portfile for OpenSMTPD's 7.4.0p0 release.

While previous minor modifications went relatively smoothly, I have run into errors using that methodology and even attempting to build from the source in the tarball using the instructions from the README.md.

I've shared the error logs here:
https://trac.macports.org/ticket/68573

For the sake of completeness, I thought that it might be worth mentioning upstream as well? My apologies in advance if this is related to a local build issue (though I did at least test to make sure that this system can build 7.3.0p2 successfully).

ryandesign:
Looks like opensmtpd is trying to declare a function `strlcpy` via its bundled copy of openbsd-compat, despite the fact that macOS already has `strlcpy`, which the configure script correctly identified.

omar-polo:
Hello,

It was caused by ebea68918a994ddaa43bb1aac769bc086f52e2f2, I forgot to update the `#ifdef` from `HAVE_DECL_STRLCPY` to `HAVE_STRLCPY` (and similarly for strlcat).  Now it should be fixed.

Thanks for the report!

omar-polo:
Can I consider this to be fixed?  We'll release a p1 in a few days, I'd like to have it working for macos too.

Thanks!

artkiver:
> Can I consider this to be fixed? We'll release a p1 in a few days, I'd like to have it working for macos too.
> 
> Thanks!

Shoot, my apologies for not testing this sooner!

I seem to be running into a different issue now? (assuming I copied the updated openbsd-compat.h locally correctly? I'm not quite a git/GitHub wizard) log attached.
[opensmtpd7.4.0p0.5.log](https://github.com/OpenSMTPD/OpenSMTPD/files/13246001/opensmtpd7.4.0p0.5.log)

omar-polo:
@artkiver 
> Shoot, my apologies for not testing this sooner!

No problem, it's my fault for hurrying :)

> I seem to be running into a different issue now?

Yes, there's some issues with how the discovery for b64_ntop is handled.  Will look into it soon.  (I should definitely add a macos machine in the CI!)

omar-polo:
7.4.0p1 is out now and includes also these fixes, so I'm going to close this issue.  Of course, feel free to reopen or file a new one if other problems arise.

Thanks!

[...]

PR #1235, opened 2023-11-08
fix res_hnok and b64_{pton,ntop} on macos

omar-polo:
since ebea68918a we're not "aggressively" depending on libresolv and this caused breakage on macos since the build there relied on the configure script adding -lresolv if found even if apparently not needed.

Now it fails to build since we include resolv.h which has #defines from res_hnok (and b64_{pton,ntop}) to the res_9_* "namespace", and not linking with -lresolve leaves those symbols undefined.

So, check additionally for the res_9_* function names to pull in -lresolv on macos.

Issue reported by @artkiver (グレェ), github issue #1232, thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -255,7 +255,14 @@ AC_SEARCH_LIBS([pidfile], [util],
 
 AC_SEARCH_LIBS([res_hnok], [resolv],
 	[AC_DEFINE([HAVE_RES_HNOK], 1, [1 if have res_hnok])],
-	[AC_LIBOBJ([res_hnok])])
+	[
+		dnl darwin (at least) has res_9_hnok plus #defines
+		dnl to "rename" it to res_hnok.
+		AC_SEARCH_LIBS([res_9_hnok], [resolv], [
+			AC_DEFINE([HAVE_RES_HNOK], 1, [1 if have res_hnok])
+			AC_LIBOBJ([res_hnok])
+		])
+	])
 
 AC_SEARCH_LIBS([res_randomid], [resolv],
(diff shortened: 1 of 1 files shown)


omar-polo:
@artkiver can you please give this a spin and see if it fixes the build issue?  Since it touches `configure.ac` please don't forget to run `./bootstrap` (or `autoreconf -vfi`) before running `./configure`.  thanks!

artkiver:
> @artkiver can you please give this a spin and see if it fixes the build issue? Since it touches `configure.ac` please don't forget to run `./bootstrap` (or `autoreconf -vfi`) before running `./configure`. thanks!

My apologies for the delay.

It is looking a lot happier with the modified configure.
[opensmtpd7.4p0.5.configure_and_make_and_make_install.log](https://github.com/OpenSMTPD/OpenSMTPD/files/13301513/opensmtpd7.4p0.5.configure_and_make_and_make_install.log)
ac (as well as the previously modified openbsd-compat.h) at least as far as building goes! Alas, at the moment I don't really have a great environment for testing SMTP traffic.

log attached for good measure.

Thank you! 🙏

omar-polo:
@artkiver thanks for testing, and also in particular for the build logs!  They saved me some time since we don't currently have a macos builder configured in the CI and I don't own a mac.

In the master branch there are also a couple of other fixes for mac too that removes various warnings at build time.  One of the things that remains to fix for mac is that `fatal()` and `fatalx()` are not considered "dead" functions (i.e. functions that never returns.)  This causes a few warnings (most of them are the "variable X is unused whenever 'if' condition is false"), but it's harmless otherwise.

PR #1237, opened 2023-11-22
Fix typo in tls error message

ryanakca:
Fix typo in tls error message


omar-polo:
Thanks!  I've forwarded the diff to the libtls maintainers, once it lands there I'll sync our boundled libtls (we're behind a few commits).

omar-polo:
Merged, thanks!

(Still need to figure it out how to make it appear as merged here on github when I manually apply the patch and push, since I wanted to keep the same commit message)

PR #1243, opened 2024-01-22
handle escapes inside quoted strings

### LINKED ISSUE #1242: Misinterpretation of escaped quotation marks in To/From/Cc
I believe I identified an issue with escaped quotation marks in From/To/Cc header fields, e.g.

To: "\\"Doe, John\\"" <john.doe@example.com>

will be misinterpreted by OpenSMTPD and the address will be modified to become

To: ""Doe@mail.example.com, John"" <john.doe@example.com>

While I do agree there is no point in formatting addresses that way and this is a bug in the mail client, the formatting is to my understanding in accordance to RFC. Hence OpenSMTPD should accept it.

omar-polo:
Fixes #1242


omar-polo:
merged manually

Issue #1252, opened 2024-06-07
Host names truncated to 64 chars


morille:
I have an OpenSMTPD 6.6.4p1 instance running on an Ubuntu 20.04.6 LTS machine, installed using package manager.
When receiving emails from hosts with a name longer than 64 characters, the name used for the fc-rDNS check is truncated to 64 chars, and the connection is discarded.
It seems that is because HOST_NAME_MAX defaults to 64 on most linux distros.
```
getconf HOST_NAME_MAX
64
```
Is there an "easy" way to change this ?

omar-polo:
Hello,

Is this a glibc-thing?  on alpine limits.h defines HOST_NAME_MAX to 255, but I can see that on debian is 64.  (the VERSION section of [gethostname(2)](https://man7.org/linux/man-pages/man2/gethostname.2.html) on linux seem have some commentary.)

I guess we'll have to use `_POSIX_HOST_NAME_MAX` which is 255 on all the systems I managed to get my hands on (OpenBSD, FreeBSD, alpine and debian)

morille:
In limits.h there is an interesting note saying that the minimum acceptable value for HOST_NAME_MAX is _POSIX_HOST_NAME_MAX.  
So when both are defined, smtpd should use the bigger one.  
Is this something you could do ? (I did not put my hands on C for almost 20 years...)

s4uliu5:
I'm using OpenSMTPD 7.5.0p0, on Funtoo Linux with glibc-2.33, with rdns and fcrdns filters enabled.

 And my system is affected by this issue.

Before applying https://github.com/morille/OpenSMTPD/commit/b9a09d61614f84e6dc9d43bd91952d85c4530d48,  I had a lot of fcrdns errors similar to this
```
Jul  1 08:56:15 smtp01 smtpd[7838]: 33c297056fc0b7f5 smtp connected address=40.107.149.96 host=mail-germanywestcentralazon11022096.outbound.protection.outlook.
Jul  1 08:56:15 smtp01 smtpd[7838]: 33c297056fc0b7f5 smtp failed-command command="" result="550 fcrDNS error"
```
as host names longer than 64 chars were truncated  that caused fcrdns filter to fail, e.g. from `mail-germanywestcentralazon11022096.outbound.protection.outlook.com` to `mail-germanywestcentralazon11022096.outbound.protection.outlook.`

d00m178:
same issue on opensmtpd-7.8.0p0

```
[...]

PR #1254, opened 2024-06-24
fix: clamp HOST_NAME_MAX

morille:
Debian based systems have an HOST_NAME_MAX lower than _POSIX_HOST_NAME_MAX and it causes the issue #1252 Clamp HOST_NAME_MAX in the sysconf way: https://manpages.ubuntu.com/manpages/trusty/fr/man3/sysconf.3.html

### DIFF
--- a/openbsd-compat/defines.h
+++ b/openbsd-compat/defines.h
@@ -38,6 +38,11 @@
 # endif
 #endif
 
+#if defined(HOST_NAME_MAX) && defined(_POSIX_HOST_NAME_MAX) && HOST_NAME_MAX < _POSIX_HOST_NAME_MAX
+# undef HOST_NAME_MAX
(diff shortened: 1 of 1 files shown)


omar-polo:
Hello,

Thanks for the PR and sorry for the delay in dealing with this issue.  Unfortunately, the PR can't go in as-is.  Changing a standard define like this in a file which is included early is very dangerous.  There are structs in header files which could use `HOST_NAME_MAX` and we can't change it from under their feet.

The way to fix this is to change the places where we use `HOST_NAME_MAX` to `_POSIX_HOST_NAME_MAX` (and eventually define the latter if not present on the system), or add a new `SMTPD_HOST_NAME_MAX` define with the right value.  Of the two, I believe the second has more chances of being committed upstream in OpenBSD too.

Issue #1106, opened 2020-12-09
Sometimes `listen on localhost` causes smtpd to not start


ryanakca:
Sometimes `listen on localhost` causes smtpd to listen on 127.0.0.1:25 twice. This reliably happens on Debian when `127.0.0.1 localhost` is included twice in `/etc/hosts`.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976194#17 for details.

PR #1256, opened 2024-08-08
Do not listen on same address multiple times

### LINKED ISSUE #1075: startup fails with "address already in use"
Debian Buster, opensmtpd 6.7.1-p1:

If /etc/hosts provides
~~~
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
ff02::1         ip6-allnodes
ff02::2         ip6-allrouters
~~~
and /etc/smtpd.conf is
~~~
listen on localhost
action "relay" relay host smtp://mailhost.example/com
match from local for any action "relay"
~~~
then smtpd fails at boot time with "address already in use". Apparently this problem does not come up if I run "systemctl restart opensmtpd" later.

Using "listen on lo" seems to avoid this problem.  There is something fishy here.

jchristgit:
When `127.0.0.1 localhost` is included twice in `/etc/hosts`, such as on
some installations of Debian, OpenSMTPD currently fails to start with
"address already in use". Fix it by only adding listeners for unique
addresses returned in a single `getaddrinfo` call.

Fixes #1075 and #1106.

### DIFF
--- a/usr.sbin/smtpd/parse.y
+++ b/usr.sbin/smtpd/parse.y
@@ -3447,9 +3447,12 @@ static int
 host_dns(struct listen_opts *lo)
 {
 	struct addrinfo		 hints, *res0, *res;
-	int			 error, cnt = 0;
-	struct sockaddr_in	*sain;
-	struct sockaddr_in6	*sin6;
+	int			error, cnt = 0;
+	int			totaladdrsv4 = 0, totaladdrsv6 = 0,
+				visitedaddrsv4 = 0, visitedaddrsv6 = 0,
+				is_configured_address;
+	struct sockaddr_in	*sain, **seenaddrsv4;
+	struct sockaddr_in6	*sin6, **seenaddrsv6;
 	struct listener		*h;
 
 	memset(&hints, 0, sizeof(hints));
@@ -3466,9 +3469,19 @@ host_dns(struct listen_opts *lo)
 	}
 
 	for (res = res0; res; res = res->ai_next) {
+		totaladdrsv4 += (res->ai_family == AF_INET);
+		totaladdrsv6 += (res->ai_family == AF_INET6);
+	}
+
+	seenaddrsv4 = xcalloc(totaladdrsv4, sizeof(struct sockaddr_in));
+	seenaddrsv6 = xcalloc(totaladdrsv6, sizeof(struct sockaddr_in6));
+
+	for (res = res0; res; res = res->ai_next) {
 		if (res->ai_family != AF_INET &&
 		    res->ai_family != AF_INET6)
 			continue;
+
+		is_configured_address = 0;
 		h = xcalloc(1, sizeof(*h));
 
 		h->ss.ss_family = res->ai_family;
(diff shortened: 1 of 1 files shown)


usr.sbin/smtpd/parse.y
```diff
@@ -3465,10 +3466,12 @@ host_dns(struct listen_opts *lo)
 		return (-1);
 	}
 
-	for (res = res0; res; res = res->ai_next) {
+	for (res = res0; res; prevaddr = res->ai_addr, res = res->ai_next) {
 		if (res->ai_family != AF_INET &&
 		    res->ai_family != AF_INET6)
 			continue;
+		if (prevaddr != NULL && memcmp(prevaddr, res->ai_addr, sizeof(struct sockaddr)) == 0)
```
omar-polo, line 3473: I believe that rather than `sizeof(struct sockaddr)` we should use `res->ai_addrlen`, i.e.

```
if (prevaddr != NULL &&
    prevaddr->ai_addrlen == res->ai_addrlen &&
    memcmp(prevaddr->ai_addr, res->ai_addr, res->ai_addrlen) == 0)
	continue;
```

Also, while here, I guess to stay on the safer side we could loop over `conf->sc_listeners` and check for duplicates there.  (imagine the case where we first add `127.0.0.1`, then `::1`, and finally `127.0.0.1` again.)

omar-polo (changes requested):
Hello,

Sorry for the delay.  I've left one nit but overall this looks fine to me, thanks!

poolpOrg:
> Testing my code, I see one problem though that I am unsure how to cleanly fix: if you configure the same listener multiple times, it will now silently just "work" which I find counterintuitive.

I have a different perspective, to me it does not silently just work, but it silently ignores failures so the end result does not necessarily work as I'd expect: maybe the address was bound on a listener and I assumed it was bound on another listener so now none of the options I expect on a specific interface are there and I may have created a dangerous config.

I don't feel too good about skipping interfaces and I think most of the problems rise from two cases:

a- the use of a wildcard address (ie: 0.0.0.0)
b- the use of a hostname as a listener

The case a- is simple to address, listeners should have their addresses resolved prior to binding and the binding should bind specific addresses first (ie: 127.0.0.1 should be bound before 0.0.0.0).

The case b- I'm not sure what to do because I think it was a mistake to allow binding on hostnames, so maybe there's a solution and we should discuss it, but it should not lead to silently ignoring stuff, it should break the daemon but be config fixable.

poolpOrg:
Thanks, I'll have a look at the diff shortly.

> Since you say you think it was a mistake to allow binding on hostnames, perhaps the default configuration should recommend to listen on lo instead? I am not sure how portable that is, though.

I don't know why the portable version suggests `listen on localhost`, either I was tired or someone else did it, but the OpenBSD version listens on an interface and in portable I'd suggest listening on IP specific addresses as I don't know if listening on an interface can be made to work in the same way as with OpenBSD.

ArchangeGabriel:
@poolpOrg This was changed in https://github.com/OpenSMTPD/OpenSMTPD/commit/8fd41496cd42592dbb0d8be1527b443139b1f881

PR #1260, opened 2024-08-20
fix: clamp HOST_NAME_MAX

morille:
Debian based systems have an HOST_NAME_MAX lower than _POSIX_HOST_NAME_MAX and it causes the issue #1252
It is usually clamped by sysconf: https://manpages.ubuntu.com/manpages/trusty/fr/man3/sysconf.3.html
Using a new SMTPD_HOST_NAME_MAX define to hold the correct value without redefining HOST_NAME_MAX

NB: This follows up another PR that I deleted by mistake: https://github.com/OpenSMTPD/OpenSMTPD/pull/1254

### DIFF
--- a/openbsd-compat/defines.h
+++ b/openbsd-compat/defines.h
@@ -32,9 +32,17 @@
 #define INFTIM	(-1)
 #endif
 
-#ifndef HOST_NAME_MAX
-# ifdef _POSIX_HOST_NAME_MAX
-# define HOST_NAME_MAX _POSIX_HOST_NAME_MAX
+#ifndef SMTPD_HOST_NAME_MAX
+# ifdef HOST_NAME_MAX
+#  if defined(_POSIX_HOST_NAME_MAX) && HOST_NAME_MAX < _POSIX_HOST_NAME_MAX
+#   define SMTPD_HOST_NAME_MAX _POSIX_HOST_NAME_MAX
+#  else
+#   define SMTPD_HOST_NAME_MAX HOST_NAME_MAX
+#  endif
+# elif defined(_POSIX_HOST_NAME_MAX)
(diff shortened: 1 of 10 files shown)


omar-polo:
Hello,

Thank you for the updated PR, I prefer this approach to changing `HOST_NAME_MAX`.  I've sent the diff (slightly tweaked so that it applies in the OpenBSD src tree) to tech@ mailing list to see what the other hackers think.

omar-polo:
Hello @it-bsi,

unfortunately I was a bit slow and this hasn't made in time for 7.6.  Some concern were raised on tech@, which are sound (but I'm not going to report them here.)  Suffice to say that after checking more carefully the situation across various systems, I still think this is the way to go.  Valid concerns notwithstanding, every one using `HOST_NAME_MAX` is already in a grey area, and changing it to `_POSIX_HOST_NAME_MAX` it's not going to make it worst (it actually is a slightly improvements.)

omar-polo:
Just wanted to apologize for having this been resting here.  There was some resistance in tech@ for getting this in, on the idea that other systems could fix their libc (something which in part I share.)

Anyway, this is just causing stress for users of portable, and if we don't run on other systems what's the deal with -portable at all?  I'm about to release 7.8.0  a bit in a rush because there's also an errata involved, and I haven't run with this in a bit so I"m not confident in just committing it before the release.

I'll take the time next week to clean this a bit and commit it though, and then sort what has to be sorted to reduce the friction with OpenBSD src.

morille:
@omar-polo There's no need to apologize, this is free software, not a public service.  
I do also not fully understand the portable version purpose if not fixing system discrepancies.  
Feel free to close this PR and work on a clean version on your side.  
And anyway thank you for trying.

PR #1275, opened 2025-02-24
fix address family typo in table.c (PF_INET->PF_INET6)

adufray:
discovered & corrected by satanist on IRC

### DIFF
--- a/usr.sbin/smtpd/table.c
+++ b/usr.sbin/smtpd/table.c
@@ -709,7 +709,7 @@ parse_sockaddr(struct sockaddr *sa, int family, const char *str)
 		if ((cp = strchr(addr, SCOPE_DELIMITER)) != NULL)
 			*cp++ = '\0';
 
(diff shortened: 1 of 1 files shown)


BetaRays:
I spent a good amount of time trying to find why my server would occasionally give temporary routing errors, finding that it was linked to IPv6, then understanding that my IPv6 wasn’t parsed properly, but someone else already found the bug. :sweat_smile: 
Interestingly, when I wanted to check on the current version of the code, I found that the bug doesn’t appear in the OpenBSD version of the code: https://cvsweb.openbsd.org/src/usr.sbin/smtpd/table.c?rev=1.54&content-type=text/x-cvsweb-markup
Anyway, thanks for fixing that annoying bug.

omar-polo:
Good catch, and apologize for not having spotted it.  I do run diff between portable and base often, but I've just missed this one :/

PR #1306, opened 2026-04-13
regress: add Perl regression test suite

nobuddytoanybuddy:
Port Alexander Bluhm's relayd-regress framework to OpenSMTPD. Each
test forks a controlled smtpd(8) process alongside an SMTP client and
an LMTP sink, drives a complete mail transaction, and asserts expected
output in the daemon log. No external test framework required — base
system Perl only (plus makemap(8) and opensmtpd-table-sqlite for the
respective backends).

Table engine coverage (all configurable backends):
- `args-table-static.pl`  — inline static table, K_VIRTUAL
- `args-table-file.pl`    — file backend, K_VIRTUAL
- `args-table-aliases.pl` — file backend, K_ALIAS
- `args-table-db.pl`      — Berkeley DB via makemap(8)
- `args-table-sqlite.pl`  — SQLite via table-sqlite(8)

Filter protocol coverage (proc-exec, v0.7):
- `args-filter-reject.pl` — mail-from phase, 550 reject
- `args-filter-junk.pl`   — mail-from phase, junk marking + delivery

All 7 tests pass against OpenBSD 7.8 / OpenSMTPD 7.6.0.

```
make SUDO=doas
make SUDO=doas SMTPD=/path/to/smtpd   # test alternative binary
```

Framework modeled on relayd-regress by Alexander Bluhm <bluhm@openbsd.org>.


omar-polo:
sorry for the long delay in handling this but let me just say: THANK YOU!

this will make me feel way more relaxed from now on :)

I'll need some time to properly read this (i like perl but i don't have the chance to read / write it often, so i'm always rusty :P)

PR #1310, opened 2026-07-06
Check that AUTH PLAIN messages are valid UTF-8

gahr:
I just got my auth table explode on an invalid UTF-8 character that was part of a password used in a brute force attack.

I think it makes sense to disallow invalid AUTH PLAIN messages as soon as possible, to avoid issues downstream, i.e., so a downstream can assume that user and pass are valid, as far as the protocol is concerned.


omar-polo:
makes sense to me, per rfc4616 those should be utf8 strings (without NUL in them).  I'll take a moment to double-check one thing about the utf8 validation but as general thing it looks good to me.

PR #1313, opened 2026-08-28
smtpd-filters: add stats and queue report subsystems

raspbeguy:
`smtpctl show stats` needs root, and it costs one imsg round trip per key. An external metrics exporter is thus not practical. The filter API sees SMTP sessions only, not the queue and not local delivery.

These commits add two report subsystems to the processor API. A `proc` declaration must enable each one. `stats` gives a periodic snapshot of the internal counters. `queue` gives the fate of every envelope. The second commit also adds counters to parts that had none.

`PROTOCOL_VERSION` stays at 0.7, so existing filters see an unchanged handshake. `IMSG_VERSION` changes to 17. You can merge the first commit alone.


omar-polo:
Hello!

in general I like this very much.  I might need a bit of time to go thru it (it's fairly big ^^;) but I'd like to have something like this available to improve observabilities :D

Just to set the right expectations though: I cannot promise that this will be done in time for 8.0, I've been a bit slow recently to handle OSS work, apologize :(, and we'll have to try to split this into smaller, more manageable commits because we very rarely commit big changes in one sweep.

PR #1314, opened 2026-09-08
table_proc: b64encode user/pass, bump protocol version

gahr:
**NOTE**: I'm opening this PR mainly for discussion. This requires a table proc protocol version bump, so if it's accepted I am willing to update the  `table-*` repos under the `OpenSMTPD` org accordingly.

The table_proc auth facility sends raw `<user>:<password>` to the backend. If either the `user` or the `password` part contains a `:`, there's no way for the backend to decide where to split the string.

I propose to base64-encode both `user` and `password` before joining them with a `:`.

This requires a protocol version bump, so tables know to base64-decode the payload.


omar-polo:
I don't see any issue with this in theory: actually, it would have prevented a bug we had due to not enough validation.  also, not many tables are using this (none of the one we provide IIRC), so it's easier to make this change.

I'll summon @poolpOrg to see what he thinks, but seems okay to me.

p.s. for the CI don't waste time, if you rebase on top of master it should pass again.

gahr:
> p.s. for the CI don't waste time, if you rebase on top of master it should pass again.

rebased

